PR #65, opened 2024-06-18
check for empty generation vals before resampling

braddf:
# Pull Request

## Description

Prevent resampling error when no generation data is available

## How Has This Been Tested?

Local API with Postman + local UI

- [x] Yes

- [x] Yes

## Checklist:

### DIFF
--- a/src/india_api/internal/service/resample.py
+++ b/src/india_api/internal/service/resample.py
@@ -5,6 +5,8 @@ from india_api.internal import (
 
 
 def resample_generation(values: list[ActualPower], internal_minutes: int) -> list[ActualPower]:
+    if not values:
+        return []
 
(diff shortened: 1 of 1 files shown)


src/india_api/internal/service/resample.py
```diff
@@ -5,6 +5,8 @@
 
 
 def resample_generation(values: list[ActualPower], internal_minutes: int) -> list[ActualPower]:
+    if not values:
```
peterdudfield, line 8: i think this should `if len(values) == 0`

peterdudfield: fair enough, then stick with that. I find it harder to read, but its probably quicker

peterdudfield (approved):

Issue #70, opened 2024-08-13
Send email with csv


peterdudfield:
Requirements:
- Send an email every dat at 8.30 IST 
- Send to multiple emails, that OCF defines
- Need to attached csv of latest wind and solar DA forecasts
- Formatting on email can be minimal

CSV file
1. Make API end point that makes csv
2. Upload (and download) csv file from S3 bucket

Who runs the cron job?
- Mail service
- Standalong micro service
- Airflow

Ideas:
- Various mail services could do it. Would be nice if the mails service has an api. 
- Full AWS solution (bad to get locked into)
- Full python code #48 

Code language?
python or node?

peterdudfield:
We could use `sendgrid`, `mailgun`, `hubspot` or `mailsend`

PR #71, opened 2024-08-14
first go at csv da endpoint

peterdudfield:
# Pull Request

## Description

First go at making a csv end point for the da forecast
This providers a csv file, rather than text that looks like a csv. This can be changed

Screenshot 2024-08-15 at 12 30 56

Screenshot 2024-08-14 at 22 55 05

The file looks like this
Screenshot 2024-08-15 at 11 29 29
* note start time would go up to 23.45

Helps with #70. 

## How Has This Been Tested?

Ran it locally

- [x] Yes

## Checklist:

### DIFF
--- a/src/india_api/internal/service/csv.py
+++ b/src/india_api/internal/service/csv.py
@@ -1,0 +1,41 @@
+import pandas as pd
+
+from india_api.internal import PredictedPower
+
+
+
+def format_csv(values: list[PredictedPower]) -> pd.DataFrame:
+    """
+    Format the predicted power values into a pandas dataframe ready for CSV export.
+
+    The pandas dataframes ends up with
+    - Date [IST]: The date
+    - Time: start and end time, e.g 00:00 to 00:15
+    - PowerKW, the forecasted power
+
+    """
+    # change list of prediction power to pandas dataframe
+    df = pd.DataFrame([y.__dict__ for y in values])
+
+    # change Time columns from UTC to IST
+    df["Time"] = pd.to_datetime(df["Time"]).dt.tz_convert("Asia/Kolkata")
+
+    # create date columns
+    df["Date [IST]"] = df["Time"].dt.date
+    # create start and end time column and only show HH:MM
+    df["Start Time [IST]"] = df["Time"].dt.strftime("%H:%M")
+    df["End Time [IST]"] = (df["Time"] + pd.to_timedelta("15T")).dt.strftime("%H:%M")
+
+    # combine start and end times
+    df["Time"] = df["Start Time [IST]"].astype(str) + " - " + df["End Time [IST]"].astype(str)
+
+    # drop start, end and Time columns and order
+    df = df.drop(columns=["Start Time [IST]", "End Time [IST]"])
+    df = df[["Date [IST]", "Time", "PowerKW"]]
+
+    # only get tomorrows results. This is for IST time.
+    now_ist = pd.Timestamp.now(tz="Asia/Kolkata")
+    tomorrow = now_ist + pd.Timedelta(days=1)
+    df = df[df["Date [IST]"] == tomorrow.date()]
+
+    return df
--- a/src/india_api/internal/service/csv.py
+++ b/src/india_api/internal/service/csv.py
@@ -29,7 +29,7 @@ def format_csv(values: list[PredictedPower]) -> pd.DataFrame:
     # combine start and end times
     df["Time"] = df["Start Time [IST]"].astype(str) + " - " + df["End Time [IST]"].astype(str)
 
-    # drop start, end and Time columns and order
+    # drop start and end columns and order
     df = df.drop(columns=["Start Time [IST]", "End Time [IST]"])
     df = df[["Date [IST]", "Time", "PowerKW"]]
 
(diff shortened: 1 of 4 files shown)


src/india_api/internal/service/csv.py
```diff
@@ -0,0 +1,41 @@
...
+    # create date columns
+    df["Date [IST]"] = df["Time"].dt.date
+    # create start and end time column and only show HH:MM
+    df["Start Time [IST]"] = df["Time"].dt.strftime("%H:%M")
+    df["End Time [IST]"] = (df["Time"] + pd.to_timedelta("15T")).dt.strftime("%H:%M")
+
+    # combine start and end times
+    df["Time"] = df["Start Time [IST]"].astype(str) + " - " + df["End Time [IST]"].astype(str)
+
+    # drop start, end and Time columns and order
```
braddf, line 32: *drop start and end time columns?

src/india_api/internal/service/server.py
```diff
@@ -133,3 +173,185 @@ class GetHealthResponse(BaseModel):
...
+            detail=f"Error getting {source} power production: {e}",
+        ) from e
+
+    return GetForecastGenerationResponse(
+        values=[y.to_timezone(tz=local_tz) for y in values],
+    )
+
+
+@server.get(
+    "/{source}/{region}/forecast/da-csv", tags=["Forecast Routes"], response_class=FileResponse
```
braddf, line 275: Wondering if the `da-csv` route segment is a bit specific, if someone wanted the full forecast, or forecast + generation, would that be extra duplicate routes? Could do just `/{source}/{region}/forecast/da-csv` with an optional param with default value like `type = "day-ahead"` or something..?

braddf: At the moment the "tomorrow only" code is in the `format_csv` function, so that might need rejigging as well if so though

braddf: So actually, the DA thing here is two things, the forecast horizon and the forecast values filter... just don't want to start using this endpoint and then almost immediately be adding another/more for other similar things

src/india_api/internal/service/server.py
```diff
@@ -1,22 +1,47 @@
...
 from india_api.internal.service.sites import router as sites_router
 from india_api.internal.service.regions import router as regions_router
-from india_api.internal.service.database_client import get_db_client
+from india_api.internal.service.database_client import get_db_client, DBClientDependency
 
 logging.basicConfig(level=logging.DEBUG, stream=sys.stdout)
 log = logging.getLogger(__name__)
 version = "0.1.32"
 
+local_tz = pytz.timezone("Asia/Kolkata")
```
peterdudfield, line 33: I need to check why this is here

src/india_api/internal/service/server.py
```diff
@@ -114,6 +140,20 @@ async def save_api_request_to_db(request: Request, call_next):
     return response
 
 
+def validate_source(source: str) -> str:
```
peterdudfield, line 143: im suprised this is here,

src/india_api/internal/service/server.py
```diff
@@ -133,3 +173,185 @@ class GetHealthResponse(BaseModel):
 def get_health_route() -> GetHealthResponse:
     """Health endpoint for the API."""
     return GetHealthResponse(status=status.HTTP_200_OK)
+
+
+class GetHistoricGenerationResponse(BaseModel):
```
peterdudfield, line 178: this seems wrong

src/india_api/internal/service/server.py
```diff
@@ -133,3 +173,185 @@ class GetHealthResponse(BaseModel):
...
+
+    # format to dataframe
+    df = format_csv(values)
+
+    # make file format
+    now_ist = pd.Timestamp.now(tz="Asia/Kolkata")
+    tomorrow_ist = df["Date [IST]"].iloc[0]
+    csv_file_path = f"{region}_{source}_da_{tomorrow_ist}.csv"
+
+    description = f"This file is made for {region} for {source} for {tomorrow_ist}, this was made at {now_ist}"
```
braddf, line 306: Maybe something like
```description = f"Generated for {region} for {source} for {tomorrow_ist}, Forecast up-to-date as of {now_ist}"```

src/india_api/internal/service/csv.py
peterdudfield: yes, thanks, ill update that

src/india_api/internal/service/server.py
peterdudfield: thanks, ill update it, not exactly this, but
`Forecast for {region} for {source} for {tomorrow_ist}. The Forecast was created at {now_ist}`

src/india_api/internal/service/server.py
peterdudfield: yea its very specific for the use case right now. 

Perhaps I can change it to `/{source}/{region}/forecast/day-ahead-csv` and then if we want to expand it to different forecast horizons, then we can

peterdudfield: or we do `/{source}/{region}/forecast/csv` and then we add forecast horizon as a parameter

I think its ok if we limit it to DA, and we can expand it later on

braddf (approved):
LGTM let's try it out 👍

Issue #81, opened 2024-09-23
Add region search too when getting sites


peterdudfield:
When loading ruvnl sites, only load ruvnl sites. Current it loads the first pv / wind site there is

https://github.com/openclimatefix/india-api/blob/main/src/india_api/internal/inputs/indiadb/client.py#L94

For the moment we got round this by making sure ad sites are second in the list, by uuid

anxkhn:
This issue seems to be resolved. Please close this issue @peterdudfield

PR #82, opened 2024-09-23
select the correct region

peterdudfield:
# Pull Request

## Description

make sure we load the correct region

#81 

## How Has This Been Tested?

CI tests,
- [x] Yes

## Checklist:

### DIFF
--- a/src/india_api/internal/inputs/indiadb/client.py
+++ b/src/india_api/internal/inputs/indiadb/client.py
@@ -89,11 +89,8 @@ class Client(internal.DatabaseInterface):
         with self._get_session() as session:
             sites = get_sites_by_country(session, country="india")
 
-            # just select wind site
-            sites = [s for s in sites if s.asset_type == asset_type]
-
-            # just select region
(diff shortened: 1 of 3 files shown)


src/india_api/internal/inputs/indiadb/client.py
```diff
@@ -91,6 +91,16 @@ def get_predicted_power_production_for_location(
 
             # just select wind site
             sites = [s for s in sites if s.asset_type == asset_type]
+
+            # just select region
```
Sukh-P, line 95: Could this be a bit simpler by adding this condition to the previous list comprehension filter? Could then have the one line:

```
# just select asset type and region
sites = [s for s in sites if s.asset_type == asset_type and s.region == location]
```

Sukh-P (approved):
LGTM, nice catch!

src/india_api/internal/inputs/indiadb/client.py
peterdudfield: thats a good idea

PR #83, opened 2024-09-24
Sentry update to the code

ADIMANV:
# Pull Request

## Description
Minor Update to the code which adds Sentry for a better error detection
Stems from the PR: https://github.com/openclimatefix/ocf-infrastructure/issues/540
## Checklist:

### DIFF
--- a/src/india_api/cmd/main.py
+++ b/src/india_api/cmd/main.py
@@ -2,11 +2,12 @@
 
 import uvicorn
 import sentry_sdk
+import version
 
 from india_api import internal
 from india_api.internal.config import Config
 from india_api.internal.service import get_db_client, server
-__version__ = "0.1.37"
+=
 
 cfg = Config()
 
@@ -17,7 +18,7 @@ sentry_sdk.init(
 )
 
 sentry_sdk.set_tag("app_name", "india_api")
-sentry_sdk.set_tag("version",__version__)
+sentry_sdk.set_tag("version",version.__version__)
 
 
 match cfg.SOURCE:
--- a/src/india_api/cmd/main.py
+++ b/src/india_api/cmd/main.py
@@ -1,6 +1,7 @@
 """The main entrypoint to the application."""
 
 import uvicorn
+import sentry_sdk
 
 from india_api import internal
 from india_api.internal.config import Config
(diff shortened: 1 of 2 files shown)


src/india_api/cmd/main.py
```diff
@@ -9,6 +10,16 @@
...
 cfg = Config()
 
+sentry_sdk.init(
+    dsn=os.getenv("SENTRY_DSN"),
+    environment=os.getenv("ENVIRONMENT", "local"),
+    traces_sample_rate=1
+)
+
+sentry_sdk.set_tag("app_name", "india_api")
+sentry_sdk.set_tag("version", india_api.__version__)
```
peterdudfield, line 20: I think you should just need to import [this](https://github.com/openclimatefix/india-api/blob/main/src/india_api/internal/service/server.py#L17)  as use this version

src/india_api/cmd/main.py
```diff
@@ -1,14 +1,25 @@
 """The main entrypoint to the application."""
 
 import uvicorn
+import sentry_sdk
 
 from india_api import internal
 from india_api.internal.config import Config
 from india_api.internal.service import get_db_client, server
-
+__version__ = "0.1.37"
```
peterdudfield, line 9: I think its better to import this, from where I said, otherwise there are two version and its hard to keep them both update to date

ADIMANV: yep, made that change

src/india_api/cmd/main.py
```diff
@@ -9,6 +11,16 @@
```
peterdudfield, line 11: I think it should be from `india_api.internal.service import get_db_client, server, version`, not `import version`

src/india_api/cmd/main.py
```diff
@@ -1,14 +1,25 @@
...
 cfg = Config()
 
+sentry_sdk.init(
+    dsn=os.getenv("SENTRY_DSN"),
+    environment=os.getenv("ENVIRONMENT", "local"),
+    traces_sample_rate=1
+)
+
+sentry_sdk.set_tag("app_name", "india_api")
+sentry_sdk.set_tag("version",version.__version__)
```
peterdudfield, line 20: becasue you have imported version, thats a string already, so you can just use `sentry_sdk.set_tag("version",version)`

peterdudfield (approved):

PR #114, opened 2025-04-04
add horizon to csv route & format func

braddf:
# Pull Request

## Description

Add forecast horizon params to CSV route to allow for `latest` and `forecast_horizon` as well as `day_ahead`.
>N.B. This will change the default behaviour of the route to be `latest`, in line with other API routes unless specified in the query parameters.

Fixes #

## How Has This Been Tested?

- [x] Test to follow next week.
- [x] _If your changes affect data processing, have you plotted any changes? i.e. have you done a quick sanity check?_

## Checklist:

### DIFF
--- a/src/india_api/internal/service/regions.py
+++ b/src/india_api/internal/service/regions.py
@@ -186,35 +186,61 @@ def get_forecast_timeseries_route(
     response_class=FileResponse,
     include_in_schema=False,
 )
-def get_forecast_da_csv(
+def get_forecast_csv(
     source: ValidSourceDependency,
     region: str,
     db: DBClientDependency,
     auth: dict = Depends(auth),
+    forecast_horizon: Optional[ForecastHorizon] = ForecastHorizon.latest,
+    forecast_horizon_minutes: Optional[int] = 0,
 ):
     """
     Route to get the day ahead forecast as a CSV file.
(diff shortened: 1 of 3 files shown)


src/india_api/internal/service/regions.py
```diff
@@ -186,35 +186,61 @@ def get_forecast_timeseries_route(
...
-    csv_file_path = f"{region}_{source}_da_{tomorrow_ist}.csv"
+    match forecast_horizon:
+        case ForecastHorizon.latest:
+            forecast_type = "intraday"
+        case ForecastHorizon.day_ahead:
+            forecast_type = "da"
+        case ForecastHorizon.horizon:
+            forecast_type = f"horizon_{forecast_horizon_minutes}"
+        case _:
+            # this shouldn't happen but will handle if class is changed
```
peterdudfield, line 238: raise 404 instead?

peterdudfield: if we remove above - we should be using `ForecastHorizon.horizon` and this should raise a `400` then

braddf: Have stripped out the `horizon` option and doubled up on the `400` exception; feels like duplication, but can't see how else the inferred types could want this if we still want to validate at the start and return early if the param isn't valid (which we do)

peterdudfield (approved):

PR #136, opened 2025-11-28
chore(repo): Refactor to sensible layout

devsjc:
Reworks the module structure of the repository to more easily allow for expansion.
Updates the way the API is configured to enable specifying what router to use at deployment.

### DIFF
--- a/src/quartz_api/internal/backends/dummydb/test_dummydb.py
+++ b/src/quartz_api/internal/backends/dummydb/test_dummydb.py
@@ -8,49 +8,49 @@ from .client import Client
 client = Client()
 
 
-class TestDummyDatabase(unittest.TestCase):
-    def test_get_predicted_wind_power_production_for_location(self) -> None:
+class TestDummyDatabase(unittest.IsolatedAsyncioTestCase):
+    async def test_get_predicted_wind_power_production_for_location(self) -> None:
         locID = "testID"
-        out = client.get_predicted_wind_power_production_for_location(locID)
+        out = await client.get_predicted_wind_power_production_for_location(locID)
         self.assertIsNotNone(out)
 
-    def test_get_predicted_solar_power_production_for_location(self) -> None:
+    async def test_get_predicted_solar_power_production_for_location(self) -> None:
         locID = "testID"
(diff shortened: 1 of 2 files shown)


src/quartz_api/cmd/main.py
```diff
@@ -1,52 +1,201 @@
...
 def run() -> None:
     """Run the API using a uvicorn server."""
+    # Get the application configuration from the environment
+    conf = ConfigFactory.parse_file((pathlib.Path(__file__).parent / "server.conf").as_posix())
+
+    # Create the FastAPI server
+    server = FastAPI(
+        version=version("quartz_api"),
+        title="Quartz API",
+        description=__doc__,
```
peterdudfield: where does __doc__ come from?

src/quartz_api/cmd/main.py
```diff
@@ -1,52 +1,201 @@
...
+    @server.get("/favicon.ico", include_in_schema=False)
+    def favicon() -> FileResponse:
+        """Serve the favicon."""
+        return FileResponse(static_dir / "favicon.ico")
+
+    @server.get("/docs", include_in_schema=False)
+    def redoc_html() -> FileResponse:
+        """Render ReDoc HTML."""
+        return FileResponse(static_dir / "redoc.html")
+
```
peterdudfield: is the /swagger endpoint still there?

src/quartz_api/internal/service/regions/__init__.py
```diff
@@ -0,0 +1,3 @@
+"""Routes for accessing regional forecasts."""
```
peterdudfield: I like this split of sites and regions

peterdudfield (approved):
its a big PR, but lots of it is just renaming and moving folders

Ive left a few comments, but I'll let address them as you need to

Quick question from me, does it run locally?

src/quartz_api/cmd/main.py
devsjc: Yeah swagger is still at the default `swagger`, as is the redoc docs at `/docs`

devsjc:
Yeah it runs fine locally!

PR #144, opened 2025-12-05
Fix Docker container ModuleNotFoundError (#143)

### LINKED ISSUE #143: docker bug:  ModuleNotFoundError: No module named 'quartz_api'
Currently when we start docker, we get the following bug

``` ModuleNotFoundError: No module named 'quartz_api'```

This is for version 0.3.4

prashantpandeygit:
The container was failing to start because the quartz_api package wasn't being installed in the virtual environment. This was due to a Python version mismatch (3.11 vs 3.12) between the build and runtime stages.
Changes:
use python 3.12 consistently across build and runtime stages and remove invalid --chown=app:app flag as user doesnt exist at runtime Fix #143

# Pull Request

## How Has This Been Tested?

1. Reproduced the original issue:
   -Built the Docker image with the original Containerfile
   - error: `ModuleNotFoundError: No module named 'quartz_api'`
   - container failed to start

2. Verified the fix
   Built the Docker image with the fixed Containerfile
   Tested the import 

## What was fixed

- pythonversion mismatch: Build stage used 3.11 while runtime used 3.12, causing path inconsistencies
- removed invalid `--chown=app:app` flag that referenced a non-existent user

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -1,4 +1,4 @@
-FROM python:3.11-slim-bookworm AS build-deps
+FROM python:3.12-slim-bookworm AS build-deps
 
 COPY --from=ghcr.io/astral-sh/uv:latest /uv /uvx /bin/
 RUN apt-get update && apt-get install -y --no-install-recommends \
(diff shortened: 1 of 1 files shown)


peterdudfield:
Thanks @prashantpandeygit

Issue #147, opened 2025-12-07
UK National API: Move into Quartz API to use Data Platform


peterdudfield:
This follows on from https://github.com/openclimatefix/client-private/issues/157

We need to move the uk-national-and-gsp api into this API

This could be done like this
- [ ] Copy routes into this API, but dont link to the database
- [ ] Make National Forecast routes
- [ ] Make National PVLive routes
- [ ] Make GSP Forecast routes
- [ ] Make GSP PVLive routes
- [ ] Make system routes
- [ ] make Hidden routes, used by UI, and airflow
- [ ] Deploy to AWS
- [ ] Add airflow checks 

When we are happy, we can 
- [ ] remake sure UI works with this API
- [ ] Change cloudflare over to this API, for api.quartz.solar (dev one first)

Some legacy things that might need to be
- [ ] Link Status with current 
- [ ] Log route to current database, so we can see what routes are being used, by which users

PR #157, opened 2025-12-17
Uk national

peterdudfield:
# Pull Request

## Description

Move uk-national over to this quartz-api

- National routes
- GSP routes
- status routes - this loads from the old database for the moment. This will change in the future
- system routes

Currently 
- forecast/all takes about ~13 seconds, which is to slow (benchmark from old app was). We currently looping over ForecastAtTimestamp
- pvlive/all takes about ~10 seconds

#147 

## How Has This Been Tested?

- [x] CI tests
- [x] ran locally, and UAT from airflow run against the API
- [x] deployed and ran on aws, 
- [x] and UAT from airflow run against the API
- [x] deploy UI to use this API - [link](https://nowcasting-q9zrv5e2n-openclimatefix.vercel.app/)

dont do in this PR: add end to end test for uk-national with the app #179 
dont do in this PR: add seperate UAT in airflow that tests it

## Checklist:


src/quartz_api/internal/service/uk_national/cache.py
```diff
@@ -0,0 +1,42 @@
...
+    params = [(k, v) for k, v in params if k not in legacy_query_params]
+
+    key = ":".join([
+        namespace,
+        request.method.lower(),
+        request.url.path,
+        repr(sorted(params)),
+    ])
+
+    log.info(f"Cache key generated: {key}")
```
peterdudfield: probably remove

src/quartz_api/internal/backends/utils.py
```diff
@@ -1,22 +1,41 @@
...
     # Window end is the beginning of the day two days ahead
-    end = (dt.datetime.now(tz=dt.UTC) + dt.timedelta(days=2)).replace(
-        hour=0,
-        minute=0,
-        second=0,
-        microsecond=0,
-    )
+    if end is None:
+        end = start + timedelta(days=4)
+
```
devsjc: I'd be tempted to get rid of this altogether and just mandate that all the timeseries routes have a start and an end

src/quartz_api/internal/service/uk_national/cache.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+async def key_builder(
+    func: Callable, # noqa
+    namespace: str = "",
+    *,
+    request: Request = None,
+    response: Response = None,   # noqa
+    args,  # noqa
+    kwargs,   # noqa
```
devsjc: Do you need all these `#noqa`'s? You have some type hinting - also you never use `args` or `kwargs`

src/quartz_api/internal/models/endpoint_types.py
```diff
@@ -106,6 +143,17 @@ class Site(SiteProperties):
...
+
+class Region(BaseModel):
+    """Region metadata."""
+
+    region_name: str = Field(..., json_schema_extra={"description": "The name of the region."})
+    region_metadata: dict | None = Field(
+        None,
+        json_schema_extra={"description": "Additional metadata about the region."},
+    )
+
```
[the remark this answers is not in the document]
peterdudfield: Could do, does LocationPropertiesBase force a lat/lon where as a Region doesnt?

src/quartz_api/internal/models/db_interface.py
```diff
@@ -31,6 +34,9 @@ async def get_predicted_solar_power_production_for_location(
         forecast_horizon_minutes: int | None = None,
         smooth_flag: bool = True,
         model_name: str | None = None,
+        start_datetime: datetime | None = None,
+        end_datetime: datetime | None = None,
+        created_utc_upper_limit: datetime | None = None,
```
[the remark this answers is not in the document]
peterdudfield: yea, im happy to change it to that

src/quartz_api/internal/backends/dataplatform/client.py
```diff
@@ -502,6 +575,133 @@ async def get_forecast_for_multiple_locations_one_timestamp(
...
+
+        # format to list of GSPYieldGroupByDatetime
+        observations_by_datetime_formated = [
+            models.GSPYieldGroupByDatetime(
+                datetime_utc=timestamp,
+                generation_kw_by_gsp_id=dict(sorted(generation_kw_by_gsp_id.items())),
+            )
+            for timestamp, generation_kw_by_gsp_id in observations_by_datetime.items()
+        ]
+        return observations_by_datetime_formated
```
devsjc: I don't think either of these functions should exist. I'm aware they might have to for this specific PR, but it goes against the desiagn of the data platform to do these massive 2D calls.

src/quartz_api/internal/backends/dummydb/client.py
```diff
@@ -219,6 +219,18 @@ async def get_substation_forecast(
 
         return values
 
+    @override
+    async def get_forecast_metadata() -> models.ForecastMetadata:
+        raise NotImplementedError()
```
devsjc: This might not be needed

src/quartz_api/internal/service/uk_national/status.py
```diff
@@ -1,27 +1,77 @@
...
+    This route is used to check the status of the last forecast run.
+    """
+    sites = await db.get_solar_regions(type="nation")
+    national_location_uuid = sites[0].region_metadata["location_uuid"]
+
+    forecast = await db.get_forecast_metadata(
+        location_uuid=national_location_uuid,
+        authdata={},
+        model_name=model_name,
+    )
```
devsjc: This should call a more generic function that just gets the latests forecasts and then filter the output of that by name. I think a DBClient function like that would be more generally useful.

src/quartz_api/internal/backends/dataplatform/client.py
```diff
@@ -318,11 +335,48 @@ async def get_substation(
...
+            reverse=True,
+        )
+        forecasts = resp.forecasts[0]
+
+        return models.ForecastMetadata(
+            initialization_timestamp_utc=forecasts.initialization_timestamp_utc,
+            created_timestamp_utc=forecasts.created_timestamp_utc,
+            forecaster_name=forecasts.forecaster.forecaster_name,
+            forecaster_version=forecasts.forecaster.forecaster_version,
+        )
```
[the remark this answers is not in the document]
devsjc: In my other comments I tried to describe that the metadata that is fetched by this route is already surfaced by the Data Platform's GetForecastAsTimeseries route. So all that has to be done is for that data to be surfaced in the `_get_solar_power_production_for_timestamp` DBClient function and it can be passed into the output of the national route without making any extra calls to the database.

devsjc:
Perhaps I can be more explicit here:

From what I can tell, there is an entire new function `get_forecast_metadata` that has been written, that calls `GetLatestForecasts` in the data platform, removes all of the returned values save the most recent, and uses the initialization timestamp of that to then enrich the returned data from the national forecast route, if the user has requested it.

I'm trying to explain that this isn't necessary, as already in the data platform's `GetForecastAsTimeseries` route, this enriched data is returned: see https://github.com/openclimatefix/data-platform/blob/eb338ab2cce40d48650e5213413546fc65ce5c11/proto/ocf/dp/dp-data.messages.proto#L80-L94. The only reason at the moment that it isn't accessible to the API response is that it is discarded in the current `get_predicted_solar_power_production_for_location` function in the Data Platform implementation of the DBClient. Modifying this function such that it also returns the metadata along with the power values is all that is needed to surface this data.

This is better because it saves a whole new database call that is, as well as being slow, potentially inaccurate to the user's requested data. They may have been returned data that isn't from the latest forecasted set, depending on their request.

src/quartz_api/internal/service/uk_national/cache.py
peterdudfield: This are actually all needed for the `cache key builder`

src/quartz_api/internal/backends/dataplatform/client.py
```diff
@@ -389,22 +447,28 @@ async def _get_predicted_power_production_for_location(
...
+                reverse=True,
+            )
+            forecaster = resp.forecasts[0].forecaster
+        else:
+            req = dp.ListForecastersRequest(forecaster_names_filter=[model_name],
+                                      latest_versions_only=True)
+            resp = await self.dp_client.list_forecasters(req, metadata={"traceid": traceid})
+            forecaster = resp.forecasters[0]
 
         req = dp.GetForecastAsTimeseriesRequest(
```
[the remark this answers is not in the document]
peterdudfield: I think this is now resovled, as i have delete the forecast_metedata route

src/quartz_api/cmd/main.py
```diff
@@ -120,6 +122,9 @@ def _create_server(conf: ConfigTree) -> FastAPI:
         },
     )
 
+    # set up cache
```
devsjc:
```suggestion
```

src/quartz_api/internal/backends/quartzdb/client.py
```diff
@@ -460,6 +463,7 @@ async def get_forecast_for_multiple_locations_one_timestamp(
         raise NotImplementedError("QuartzDB client does not support multi-location forecasts.")
 
 
+
```
devsjc:
```suggestion
```

src/quartz_api/internal/models/endpoint_types.py
```diff
@@ -39,12 +68,12 @@ def to_timezone(self, tz: str) -> "PredictedPower":
             created_time=self.created_time.astimezone(tz=ZoneInfo(key=tz)),
         )
 
-
 class ActualPower(BaseModel):
     """Defines the data structure for an actual power value returned by the API."""
 
     PowerKW: float
     Time: dt.datetime
+    location_uuid: str = Field("not-set", exclude=True)
```
devsjc: I like this being a UUID and then pydantic can validate it at the edge.

src/quartz_api/internal/service/uk_national/gsp.py
```diff
@@ -18,10 +32,16 @@
...
+@cache(key_builder=key_builder)
 async def get_forecasts_for_a_specific_gsp(
     db: DBClientDependency,
     auth: AuthDependency,
-) -> Forecast | list[ForecastValue]:
+    gsp_id: int,
+    forecast_horizon_minutes: int | None = None,
+    start_datetime_utc: str | None = None,
+    end_datetime_utc: str | None = None,
+    creation_utc_limit: str | None = None,
```
devsjc: The defaults should go here now. Also, I think the datetimes should be datetime objects so the pydantic/fastapi can do the validation for us at the top of the request.

```suggestion
    forecast_horizon_minutes: int = 0,
    start_datetime_utc: dt.datetime = (dt.datetime.now(tz=dt.UTC) - dt.timedelta(days=2)).replace(hour=0, minute=0, second=0, microsecond=0)
    end_datetime_utc: dt.datetime = (dt.datetime.now(tz=dt.UTC) + dt.timedelta(days=2)).replace(hour=0, minute=0, second=0, microsecond=0)
    creation_utc_limit: dt.datetime = dt.datetime.now(tz=dt.UTC),
```

src/quartz_api/internal/service/uk_national/gsp.py
```diff
@@ -74,9 +138,242 @@ async def get_truths_for_a_specific_gsp(
     Only 3 days of history is available. If you want to get more PVLive data,
     please use the [PVLive API](https://www.solar.sheffield.ac.uk/api/)
     """
-    raise NotImplementedError()
+    start_datetime_utc = format_datetime(start_datetime_utc)
+    end_datetime_utc = format_datetime(end_datetime_utc)
```
devsjc: As above, can go.

src/quartz_api/internal/service/uk_national/gsp.py
```diff
@@ -74,9 +138,242 @@ async def get_truths_for_a_specific_gsp(
...
+
+    start_datetime_utc = format_datetime(start_datetime_utc)
+    end_datetime_utc = format_datetime(end_datetime_utc)
+    creation_limit_utc = format_datetime(creation_limit_utc)
+
+    if start_datetime_utc is not None:
+        start_datetime_utc = ceil_30_minutes_dt(start_datetime_utc)
+    if end_datetime_utc is not None:
+        end_datetime_utc = floor_30_minutes_dt(end_datetime_utc)
+
```
devsjc: This logic could go into the route validation as well I suspect.

src/quartz_api/internal/service/uk_national/gsp.py
```diff
@@ -74,9 +138,242 @@ async def get_truths_for_a_specific_gsp(
...
+    # TODO
+    # end_datetime_utc = limit_end_datetime_by_permissions(permissions, end_datetime_utc)
```
devsjc: Is this going to be done in this PR?

src/quartz_api/internal/service/uk_national/gsp.py
```diff
@@ -74,9 +138,242 @@ async def get_truths_for_a_specific_gsp(
...
+    include_in_schema=False,
+)
+@cache(key_builder=key_builder, expire=60*30)
+async def get_truths_for_all_gsps(
+    db: DBClientDependency,
+    auth: AuthDependency,  # noqa
+    regime: str = "in-day",
+    start_datetime_utc: str | None = None,
+    end_datetime_utc: str | None = None,
+    gsp_ids: str | None = None,
```
devsjc: You know the drill here!

src/quartz_api/internal/service/uk_national/national.py
devsjc: The `ModelName` enum class here should be renamed `ForecasterName` (and potentially be moved to the `models.py` module)

src/quartz_api/internal/service/uk_national/national.py
```diff
@@ -39,10 +51,10 @@ class ModelName(str, Enum):
...
     status_code=status.HTTP_200_OK,
 )
+@cache(key_builder=key_builder)
 async def get_national_forecast(
     db: DBClientDependency,
     auth: AuthDependency,
-    request: Request,
     forecast_horizon_minutes: int | None = None,
     include_metadata: bool = False,
     start_datetime_utc: str | None = None,
```
devsjc: Again, set remove optional and set defaults here

src/quartz_api/internal/service/uk_national/status.py
```diff
@@ -1,27 +1,82 @@
...
+    # Get the national forecast,
+    # but just get it for one datestamp (to make it quick)
+    predict_powers = await db.get_predicted_solar_power_production_for_location(
+        location_uuid=national_location_uuid,
+        authdata={},
+        model_name=model_name,
+        start_datetime=start_datetime,
+        end_datetime=start_datetime,
+    )
+
```
devsjc: Be way quicker to use GetLatestForecasts here. Seems very wasteful to pull a forecast just to throw it all away to get the runtime.

src/quartz_api/internal/service/uk_national/test_time_utils.py
devsjc: Most of this can go

devsjc (changes requested):
Lots of notes to do with changing where default values are managed and removing a bunch of optional ones. I think this makes the logic path much easier to follow.

src/quartz_api/internal/backends/dataplatform/client.py
```diff
@@ -36,14 +38,22 @@ async def get_predicted_solar_power_production_for_location(
         forecast_horizon: models.ForecastHorizon = models.ForecastHorizon.latest,
         forecast_horizon_minutes: int | None = None,
         smooth_flag: bool = True,
+        forecaster_name: str | None = None,
+        start_datetime: dt.datetime | None = None,
+        end_datetime: dt.datetime | None = None,
+        created_before_datetime: dt.datetime | None = None,
```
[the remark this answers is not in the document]
peterdudfield: I agree with datetimes, but the `forecaster_name` sometimes we just want to pull a forecast, and API doesnt need to know which model to pull this for.

src/quartz_api/internal/service/uk_national/gsp.py
peterdudfield: there's a slightly complicated thing of rouding them to the low 6 hours, so I think i prefer them in a seerpate functions

src/quartz_api/internal/service/uk_national/gsp.py
```diff
@@ -43,16 +63,60 @@ async def get_forecasts_for_a_specific_gsp(
     - **creation_utc_limit**: optional, only return forecasts made before this datetime.
     returns the latest forecast made 60 minutes before the target time)
     """
-    raise NotImplementedError()
+    start_datetime_utc = format_datetime(start_datetime_utc)
+    end_datetime_utc = format_datetime(end_datetime_utc)
+    creation_utc_limit = format_datetime(creation_utc_limit)
```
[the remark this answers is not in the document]
peterdudfield: its still needed, as people enter strings of datetimes, that need to be formatted

src/quartz_api/internal/service/uk_national/gsp.py
peterdudfield: I think in general I prefer to keep the API interface exactly the same, during this phase

src/quartz_api/internal/service/uk_national/time_utils.py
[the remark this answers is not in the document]
peterdudfield: I think this needs to stay. We need to have a function somewhere that floor the start time to the nearest 6 hours, and then minuses it. I think this is the right place for it.

src/quartz_api/internal/backends/dataplatform/client.py
```diff
@@ -379,7 +402,7 @@ async def _get_predicted_power_production_for_location(
                 traceid,
             )
 
-        start, end = get_window()
+        start, end = get_window(start=start_datetime, end=end_datetime)
```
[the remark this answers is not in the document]
peterdudfield: Ive rolled the changes back, and actually left this in there. I think for some routes, like get_substation_forecast, I would prefer to use it. Also the other backend use it.

src/quartz_api/internal/service/uk_national/national.py
```diff
@@ -1,16 +1,28 @@
 """The 'national' FastAPI router object."""
 
+from datetime import UTC, datetime
```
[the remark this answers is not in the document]
peterdudfield: yep good point, ill for `import datetime as dt`

src/quartz_api/internal/backends/utils.py
```diff
@@ -3,20 +3,25 @@
...
+    if end is None:
+        end = (dt.datetime.now(tz=dt.UTC) + dt.timedelta(days=2)).replace(
+            hour=0,
+            minute=0,
+            second=0,
         microsecond=0,
     )
+
     return (start, end)
+
```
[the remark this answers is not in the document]
devsjc: That almost sounds worse, having a mixture of the two. I think we should either do it everywhere or nowhere, and for me, the benefits of doing it are far outweighted by the negatives. Here's how it seems to me:

Pros:
- Doesn't change the API signature
- Signals to the user of the API what the defaults are in the documentation, making things clearer and providing an example datetime string
- Signals to code maintainers what the defaults are for each route, enables different defaults on different routes
- Automatic validation and type casting via pydantic from the input strings
- Gets rid of a whole file and test file; cleaner code path

Cons:
- Need to slightly modify logic in existing routers; so a bit more work
- Duplication of defaults across routes

For me, the pros outweigh the cons.

src/quartz_api/internal/service/uk_national/national.py
devsjc: Ahh I see

src/quartz_api/internal/service/uk_national/system.py
```diff
@@ -30,4 +32,41 @@ async def get_system_details(
...
+
+        if gsp_id is not None and gsp_id != location.gsp_id:
+            continue
+
+        if gsp_id is not None and gsp_id == location.gsp_id:
+            return [location]
+
+        locations.append(location)
+
+    # sort by gsp_id
```
devsjc:
```suggestion
```

src/quartz_api/internal/service/uk_national/test_time_utils.py
peterdudfield: 1. Its not pandas
2. the logic of floor 6 and minus 2 days I think is worth its own function
3. linting didnt let me put datetime.now() in the defaults
```B008 Do not perform function call `dt.datetime.now` in argument defaults; instead, perform the call within the function, or read the default from a module-level singleton variable```

src/quartz_api/internal/backends/utils.py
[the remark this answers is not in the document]
peterdudfield: I think you've convienced me with pros and cons. Lets do it

PR #183, opened 2026-01-22
Fix rounding before scaling in substation forecast route

### LINKED ISSUE #182: Bug: Substation
## Describe the bug
Two different routes proces slightly different forecast results

## To Reproduce
1. pull forecasts for on substation
2. pull forecast for all substations at one timestamp

## Expected behavior
They results for one substation and one timestamp should be the same

## Additional context
- the pull forecast subation, gets the gsp data, rounds it, then scales to primary
- the all susbation routes, gets all the gsp forecast and then scales (no rounding)

## Idea
- We should remove the round before it scales to primary substation ([here](https://github.com/openclimatefix/quartz-api/blob/main/src/quartz_api/internal/backends/dataplatform/client.py#L493))
- round both routes to 3 d.c.p

CodeVishal-17:
## Description

Fixes #182

Removes rounding during GSP → substation scaling in the
`/substations/forecast` route so that both substation forecast routes
follow the same order of operations.

Previously, one route rounded values before scaling while another did not,
leading to small but inconsistent forecast values for the same substation
and timestamp. This change ensures scaling is performed first and rounding
is applied consistently at the response/model layer.

## How Has This Been Tested?

- Tests were attempted locally but require FastAPI and related dependencies
  which were not available in the local environment.
- CI is expected to run the full test suite in the correct environment.

- [x] Yes

_If your changes affect data processing, have you plotted any changes?_

- [ ] Yes (not applicable – deterministic one-line numerical fix)

## Checklist:

- [x] My code follows OCF's coding style guidelines
- [ ] I have made corresponding changes to the documentation (not required)
- [ ] I have added tests that prove my fix is effective (covered by CI)


CodeVishal-17:
Thanks for the suggestion 👍

I’ve updated the substation forecast to explicitly round values *after*
scaling (to 2 decimal places), so the order of operations is now consistent
across both substation forecast routes.

PR #214, opened 2026-03-20
fix(dpclient): Correct watts -> kilowatts

devsjc:
Fixes a bug where kilowatt values were being converted incorrectly. Also hardcodes the blend forecaster version for speed.

### DIFF
--- a/src/quartz_api/internal/service/uk_national/gsp_router.py
+++ b/src/quartz_api/internal/service/uk_national/gsp_router.py
@@ -32,6 +32,9 @@ if TYPE_CHECKING:
 
 log = logging.getLogger(__name__)
 
+GSP_FORECASTER_NAME = "blend"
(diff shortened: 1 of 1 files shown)


src/quartz_api/internal/service/uk_national/gsp_router.py
```diff
@@ -114,14 +114,17 @@ async def get_forecasts_for_a_specific_gsp(
         created_cutoff=creation_utc_limit,
         forecast_horizon_minutes=forecast_horizon_minutes or 0,
         forecaster_name="blend",
-        forecaster_version=None,
+        forecaster_version="1.3.0",
```
peterdudfield: im slightly worried this will get losts. Could you put it in a constanst file, or move this hard code tot he top of this file?

PR #272, opened 2026-04-01
fix(cache): force keys to lowercase

braddf:
# Pull Request

## Description

Forces all cache keys to lowercase by calling .lower() on the key_builder output, ensuring cache hits regardless of the case of query parameter values passed by the client. 
Existing hardcoded warm keys are already lowercase so read/write consistency is maintained. 
The @cache decorator uses key_builder for both lookup and store, so both paths are covered by this change.

Fixes #

## How Has This Been Tested?

- [x] Locally

- [x] Yes

## Checklist:


peterdudfield:
I wonder if the tests are failing becasue the timezon 'Z' got moved down to a `z`, this might have to be updated in the test

braddf:
Yeah I think def something like that with the datetime strings – right now we're only aware of the boolean attributes that are allowed to be in whatever case, so have just made it more specific for now.

devsjc (approved):

src/quartz_api/internal/service/uk_national/cache.py
```diff
@@ -50,7 +50,7 @@ async def key_builder(
             repr(sorted(params)),
             repr(sorted(permissions)),
         ],
-    )
+    ).replace("False", "false").replace("True", "true")
```
peterdudfield, line 53: couldnt you do lower? 

And then I think in one of the tests you need to use t (not T) and z( not Z)?

braddf: I think it's actually explicitly boolean params that we're worried about, so I'd rather not mess with datestamp params in case of any odd side effects it could cause...

braddf: Unless you can think of any other params that would also need to be case-insensitive @peterdudfield?

peterdudfield: ah i see, hmmm, yea, maybe just stick with this for the moment.

peterdudfield (approved):
Looks good to go, 

There might be a edge case where users use FALSE, but perhaps we dont need to worry about this for the moment

PR #274, opened 2026-04-02
feat(backends): Swap to std python grpc lib

devsjc:
Modifies the GRPC client to use the default python GRPC library.

This is built on C, and doesn't do the same Serde setup of betterproto so should be much lighter on memory. As such, we shouldn't need the sync client any more. 

I've also modified the tests such that most of the functionality is tested in unit tests, rather than integration tests. This is faster and cleaner.

Testing in development showed a much flatter memory profile and no significant change to CPU. In the below graphs, anything before the 14:30 in the middle was the old library, and afterwards, the new one:

image


pyproject.toml
```diff
@@ -19,7 +19,7 @@ authors = [
 classifiers = ["Programming Language :: Python :: 3"]
 dependencies = [
     "cryptography >= 41.0.0",
-    "dp-sdk",
+    "dp-sdk >= 0.25.0",
```
peterdudfield, line 22: should pin

src/quartz_api/cmd/main.py
```diff
@@ -99,16 +93,15 @@ async def _lifespan(server: FastAPI, conf: ConfigTree) -> AsyncGenerator[None]:
             storage = DummyStorage()
             log.warning("disabled backend. NOT recommended for production")
         case "dataplatform":
-            grpc_channel = EnrichedChannel(
-                host=conf.get_string("backend.dataplatform.host"),
-                port=conf.get_int("backend.dataplatform.port"),
+            from ocf.dp.dp_data import service_pb2_grpc
```
peterdudfield, line 96: im never quite sure on this, should imports juts be at the top, I get it that you might not need to import it. Probably a balance of readability vs speed?

src/quartz_api/cmd/main.py
```diff
@@ -132,7 +125,6 @@ async def _lifespan(server: FastAPI, conf: ConfigTree) -> AsyncGenerator[None]:
 
     server.dependency_overrides[models.get_storage_client] = lambda: storage
 
-    warm_task = None
```
peterdudfield, line 135: do you need this, incase youa re not running `uk_national`

src/quartz_api/internal/backends/dataplatform/client.py
```diff
@@ -184,69 +122,71 @@ async def get_predicted_generation(
...
-                    ),
-                    "p90": int(
-                        float(v["effective_capacity_watts"]) \
-                            * float(v["other_statistics_fractions"].get("p90", 0)) / 1000.0,
+            # Spoof the forecast values so that the capacity and id corresponds to the substation
+            resp.location_uuid = str(location_uuid)
+            for v in resp.values:
+                v.effective_capacity_watts = location.effective_capacity_watts
+
+        def _map_resp(resp: messages_pb2.GetForecastAsTimeseriesResponse) \
```
peterdudfield, line 157: should we have nested definitions? I wonder if they don't encourage testing and I think they should be moved out. What do you think? Same as below.

some reading - https://realpython.com/inner-functions-what-are-they-good-for/#using-inner-vs-non-public-helper-functions

src/quartz_api/tests/integration/uk_national/test_app.py
```diff
@@ -1,480 +1,17 @@
...
-#     """Test a sample endpoint for UK National forecast data."""
-
-#     response = await api_client_uk_national.get("/v0/solar/GB/status")
-#     assert response.status_code == 200
-#     data = response.json()
-#     assert isinstance(data, list)
-#     assert len(data) == 11
-
-
-# 6.2 check status routes
```
peterdudfield, line 452: where did these tests go?

src/quartz_api/tests/integration/uk_national/test_app.py
```diff
@@ -1,480 +1,17 @@
...
-
-    # Calling with specified gsps returns all timestamps for those gsps
-    response = await api_client_uk_national.get("/v0/solar/GB/gsp/pvlive/all?gsp_ids=1,2,3")
-    assert response.status_code == 200
-    data = response.json()
-    assert isinstance(data, list)
-    assert len(data) == 10
-
-
-# 5.1 check system routes
```
peterdudfield, line 421: where did this test go?

peterdudfield (commented):
Ive left a few comments. 

And it looks like we want to run it on dev for a few days, to make sure
1. there are no errors
2. the memory and cpu are ok

peterdudfield:
Looks like there are some errors at the moment

Screenshot 2026-04-17 at 12 19 24

pyproject.toml
devsjc: Because of the requirement to put a wheel in explicitly, and because of the backwards compatibility of the proto going forwards, I actually don't think it's much of a problem really here. But I can pin it to prevent a major bump.

peterdudfield: I think it might be wroth it, thanks

src/quartz_api/internal/backends/dataplatform/client.py
devsjc: I don't think it's necessary - these mapping inner functions are specific to the outer function they exist in, and are only going to be tested in that context. Agree that normally you would want to keep things atomic like that but here the claim is made that they are part of that outer function really, and only seperatly defined to allow specification of running on the threadpool.

src/quartz_api/tests/integration/uk_national/test_app.py
devsjc: Into a new file in the uk_national folder!

devsjc:
More stats from APItally showing a decreased memory useage:

image

src/quartz_api/tests/integration/uk_national/test_app.py
devsjc: Into a new file in the uk_national folder!

peterdudfield (approved):

PR #303, opened 2026-04-29
make auth more robust, if not email key

peterdudfield:
# Pull Request

## Description

Fix if the authentication has no email key. This can happen if auth is done via secret id and key. 

Helps with https://github.com/openclimatefix/client-private/issues/418

## How Has This Been Tested?

- CI tests

## Checklist:

### DIFF
--- a/src/quartz_api/internal/middleware/auth.py
+++ b/src/quartz_api/internal/middleware/auth.py
@@ -86,6 +86,8 @@ class AuthClient:
             if EMAIL_KEY in claims:
                 email = claims[EMAIL_KEY]
                 set_consumer(request, identifier=email)
+            else:
+                log.warning("Authentication passed but could not find email claim in JWT")
 
             return claims
 
--- a/src/quartz_api/internal/middleware/auth.py
+++ b/src/quartz_api/internal/middleware/auth.py
@@ -83,8 +83,9 @@ class AuthClient:
                 raise e
 
             # set the apitally consumer, this means we can see who has used which routes
-            email = claims["https://openclimatefix.org/email"]
-            set_consumer(request,identifier=email)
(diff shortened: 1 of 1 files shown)


src/quartz_api/internal/middleware/auth.py
```diff
@@ -83,8 +83,9 @@ async def _proxy_dependency(
                 raise e
 
             # set the apitally consumer, this means we can see who has used which routes
-            email = claims["https://openclimatefix.org/email"]
-            set_consumer(request,identifier=email)
+            if EMAIL_KEY in claims:
```
braddf, line 86: What is `EMAIL_KEY` / where does it get set..?

peterdudfield: https://github.com/openclimatefix/quartz-api/blob/main/src/quartz_api/internal/middleware/auth.py#L13

peterdudfield: Its "https://openclimatefix.org/email" and just set at the top of this file

braddf: Righhht okay, we just weren't using it before, got it

braddf (approved):
Would be good to be able to tell if we are connecting ourselves using client_id/client_secret, but at least this will stop the errors 👍

peterdudfield:
> Would be good to be able to tell if we are connecting ourselves using client_id/client_secret, but at least this will stop the errors 👍

I could add a log warning? If you think that would be useful

peterdudfield:
> > Would be good to be able to tell if we are connecting ourselves using client_id/client_secret, but at least this will stop the errors 👍
> 
> I could add a log warning? If you think that would be useful

I slipped this in so we can see whats happen

PR #312, opened 2026-05-19
v1 NL final tweaks

### LINKED ISSUE #310: V1 Quartz Api NL snag list
Following on from #199  and #309 

## Changes 15/05/26

- [x] Add auth0 callback
- [x] Get forecast working for zeeland (https://api.quartz.solar/v1/NL/solar/regions/zeeland/forecast)
- [x] forecast last update doesnt work for zeeland
- [x] /forecast/snapshot doesnt have any values
- [x] remove License in description, or discuss internally
- [x] remove blend_adjust from NL provinces
- [x] support emails incorrect/inconsistent
- [x] hide dark mode toggle
- [x] /period pre-warmed endpoints wrong default model

## Changes 18/05/26

- [x] Refresh hooks in Airflow for cache refreshing - https://github.com/openclimatefix/airflow-dags/pull/653
- [ ] Forecast metadata(?) select / ensure not misleading
  - [x] Updated names for now, can revisit "verbose" forecast values later
- [x] ocf.energy domain setup
- [x] `horizon_minutes` null in forecast response bug
- [x] Period timestamps latest created/init not earliest (bug)

Not urgent
- [ ] 15 or 30 mins floor?
- [ ] remove `delete_me` locations
- [ ] refactor tests in seprate files
- [ ] add integration tests for specific DP etc. functions
- [ ] refactor any duplicate functionality into reusable utilities
- [ ] move region helpers to RegionType as class methods where poss
- [ ] test GB (first focus is on NL)
- [ ] Finalise rate-limiting

braddf:
# Pull Request

## Description

- **NL province period endpoints** — region name filter was matching against internal DP names instead of display names, so province filters returned no results
- **Fix naming mismatch** — use `nednl` everywhere instead of mix of `ned_nl`
- **`last_updated_utc` / `latest_init_utc` in `/forecasts/period`** — cache warm was capturing timestamps from the first pgv seen; now takes the max across all regions so metadata reflects the most recent run, not the earliest
- **Rename `created_utc` → `last_updated_utc`, `init_utc` → `latest_init_utc`** — old names implied these described all values in the stitched series, not just the latest run
- **Rename `forecast_horizon_minutes` → `horizon_minutes`** — shorter; omitted from response when not set rather than returning `null`
- **NL uncurtailed forecast models** — added `ecmwf_mo_sat_uncurtailed` / `ecmwf_mo_sat_uncurtailed_adjust` slugs for `nl_regional_ecmwf_mo_sat` to NL national and regional model lists (unadjusted-only for regional)
- **Remove cache TTL disclosures** from v1 route docstrings
- **Inline `ForecastModel` in `FM` class** — remove `_m` shorthand helper
- **Rename `nednl` observer → `ned_nl`** — user-facing API name now uses underscore convention; internal DP observer name (`nednl`) is preserved via a new `slug` field on `GenerationSource`, consistent with the existing `ForecastModel` pattern
- **Rename `FixedNLProvinceClient` → `NLProvinceClient`** in tests

Fixes #310 (again)

## How Has This Been Tested?

- [x] Locally (manual checks)
- [x] Unit tests (local / CI)

- [x] Yes

## Checklist:

### DIFF
--- a/src/quartz_api/internal/service/v1/cache.py
+++ b/src/quartz_api/internal/service/v1/cache.py
@@ -154,7 +154,7 @@ async def warm_v1_forecast_cache(
             energy_type.name.lower(),
             region_type,
         )
-        any_pgv = None
+        first_found_pgv = None
         last_updated: dt.datetime | None = None
         latest_init: dt.datetime | None = None
 
@@ -169,8 +169,8 @@ async def warm_v1_forecast_cache(
                 forecaster_name=rt.default_model,
             )
             if pgvs:
-                if any_pgv is None:
-                    any_pgv = pgvs[0]
+                if first_found_pgv is None:
+                    first_found_pgv = pgvs[0]
                 for pgv in pgvs:
                     if pgv.created_timestamp and (
                         last_updated is None or pgv.created_timestamp > last_updated
@@ -203,10 +203,10 @@ async def warm_v1_forecast_cache(
 
         meta = {
             "model_name": internal_to_api_name(
-                any_pgv.forecaster_name if any_pgv else None,
+                first_found_pgv.forecaster_name if first_found_pgv else None,
                 rt,
             ),
-            "model_version": any_pgv.forecaster_version if any_pgv else None,
+            "model_version": first_found_pgv.forecaster_version if first_found_pgv else None,
             "last_updated_utc": last_updated.isoformat() if last_updated else None,
             "latest_init_utc": latest_init.isoformat() if latest_init else None,
             "cache_updated_utc": dt.datetime.now(tz=dt.UTC).isoformat(),
--- a/src/quartz_api/internal/service/v1/cache.py
+++ b/src/quartz_api/internal/service/v1/cache.py
@@ -154,7 +154,9 @@ async def warm_v1_forecast_cache(
             energy_type.name.lower(),
             region_type,
         )
-        first_pgv = None
+        any_pgv = None
+        last_updated: dt.datetime | None = None
+        latest_init: dt.datetime | None = None
 
         for i, region in enumerate(regions):
             pgvs = await db.get_predicted_generation(
@@ -166,8 +168,18 @@ async def warm_v1_forecast_cache(
                 authdata={},
                 forecaster_name=rt.default_model,
             )
-            if pgvs and first_pgv is None:
-                first_pgv = pgvs[0]
+            if pgvs:
+                if any_pgv is None:
+                    any_pgv = pgvs[0]
+                for pgv in pgvs:
+                    if pgv.created_timestamp and (
+                        last_updated is None or pgv.created_timestamp > last_updated
+                    ):
+                        last_updated = pgv.created_timestamp
+                    if pgv.init_timestamp and (
+                        latest_init is None or pgv.init_timestamp > latest_init
+                    ):
+                        latest_init = pgv.init_timestamp
             values = [
                 {
                     "time": v.valid_timestamp.isoformat(),
@@ -189,16 +201,14 @@ async def warm_v1_forecast_cache(
             # yield to event loop; keeps live requests responsive during warm
             await asyncio.sleep(0.1)
 
-        created = first_pgv.created_timestamp if first_pgv else None
-        init = first_pgv.init_timestamp if first_pgv else None
         meta = {
             "model_name": internal_to_api_name(
-                first_pgv.forecaster_name if first_pgv else None,
+                any_pgv.forecaster_name if any_pgv else None,
                 rt,
             ),
-            "model_version": first_pgv.forecaster_version if first_pgv else None,
-            "created_utc": created.isoformat() if created else None,
-            "init_utc": init.isoformat() if init else None,
+            "model_version": any_pgv.forecaster_version if any_pgv else None,
+            "last_updated_utc": last_updated.isoformat() if last_updated else None,
+            "latest_init_utc": latest_init.isoformat() if latest_init else None,
             "cache_updated_utc": dt.datetime.now(tz=dt.UTC).isoformat(),
         }
         await backend.set(f"{base}:_meta", json.dumps(meta), expire=86400)
--- a/src/quartz_api/internal/service/v1/country_config.py
+++ b/src/quartz_api/internal/service/v1/country_config.py
@@ -338,7 +338,10 @@ COUNTRIES: dict[str, CountryConfig] = {
         ),
         generation_sources=(
             GenerationSource(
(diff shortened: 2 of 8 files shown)


src/quartz_api/internal/service/v1/routes/forecasts.py
```diff
@@ -57,6 +57,7 @@
     "/{country}/{source}/regions/{region}/forecast",
     status_code=status.HTTP_200_OK,
     response_model=ForecastResponse,
+    response_model_exclude_none=True,
```
peterdudfield, line 60: I cant see this on the current api / what does this bit do?

src/quartz_api/internal/service/v1/test_router.py
```diff
@@ -254,6 +254,48 @@ async def trial_client() -> AsyncGenerator[AsyncClient, None]:
         yield ac
 
 
+_FIXED_NL_PROVINCE_UUID = uuid4()
+_NL_PROVINCE_INTERNAL_NAME = "nl_region_2_friesland"
+_NL_PROVINCE_DISPLAY_NAME = "friesland"
+
+
+class FixedNLProvinceClient(NationNameStorageClient):
```
peterdudfield: Why does this class have `Fixed` in the name?

peterdudfield: I see now becasue the uuid is fixed, I got confused becasue it was in the tests.

Maybe just call it `NLProvinceClient` and the uuid being fixed is a detail

src/quartz_api/internal/service/v1/routes/forecasts.py
braddf: This means that if the property isn't set on the response, i.e. is `null`, it's not included – so that the default forecast horizon is left unset if the user doesn't provide it. Thought this might be preferable to defaulting to `horizon_minutes: 0` but open to thoughts

src/quartz_api/internal/service/v1/test_router.py
```diff
@@ -1686,3 +1728,71 @@ def test_country_config_intraday_default_in_intraday_models() -> None:
...
+                    "power_kW": 100.0,
+                    "plevels_kW": {},
+                },
+            ],
+        ).encode(),
+        expire=3600,
+    )
+
+    # Filter by display name — should match despite internal name being different.
+    resp = await nl_province_client.get(
```
peterdudfield, line 1785: to make this test simplier, could you just set up things in the data-platform, and then hit this route twice, 1. starts putting data in cahce, then wait 1 or 2 seconds, then hit the route again and check it works

src/quartz_api/internal/service/v1/test_router.py
braddf: Fair I can rename, I think I might have had two similar clients in here previously but simplified into just this one and the standard classes

src/quartz_api/internal/service/v1/test_router.py
peterdudfield: (maybe thats not simplier)

braddf: v1 doesn't have any integration tests yet, I've focused on getting a solid unit test foundation in first to try to test things in an "isolated" way, esp. with things like the cache that is separate implementation logic to what we're testing here, which is whether the name filter function(s) are working as expected... the cache is linked in this case because the route itself relies on it by design, so we do have to minimally mock that functionality

src/quartz_api/internal/service/v1/cache.py
```diff
@@ -154,7 +154,9 @@ async def warm_v1_forecast_cache(
             energy_type.name.lower(),
             region_type,
         )
-        first_pgv = None
+        any_pgv = None
```
peterdudfield: I think i prefer `first_pgv`, rather than `any_pgv`, any feels a bit ambigious, but first gives a bit of a clue to what is used

src/quartz_api/internal/service/v1/country_config.py
```diff
@@ -139,46 +152,69 @@ class FM:
...
+    NL_UNCURTAILED = ForecastModel(
+        name="nl_regional_ecmwf_mo_sat",
+        label="ECMWF + Met Office + Satellite, Uncurtailed",
+        slug="ecmwf_mo_sat_uncurtailed",
+    )
+    NL_UNCURTAILED_ADJUST = ForecastModel(
+        name="nl_regional_ecmwf_mo_sat_adjust",
+        label="ECMWF + Met Office + Satellite, Uncurtailed (Trend Adjusted)",
+        slug="ecmwf_mo_sat_uncurtailed_adjust",
+    )
```
peterdudfield, line 217: Looks good

src/quartz_api/internal/service/v1/test_router.py
peterdudfield: Ah all ok. 

Would you be able to make a issue for making integrations tests for v1 please? Then we can close this comment

src/quartz_api/internal/service/v1/routes/forecasts.py
peterdudfield: I'd probably just leave it as `null` for the moment. Then we can have a look at this separately. There may be some other routes we want to do this for? What do you think?

src/quartz_api/internal/service/v1/cache.py
```diff
@@ -154,7 +154,9 @@ async def warm_v1_forecast_cache(
             energy_type.name.lower(),
             region_type,
         )
-        first_pgv = None
+        first_found_pgv = None
```
peterdudfield, line 157: thanks

src/quartz_api/internal/service/v1/test_router.py
peterdudfield: or add to your snag list

src/quartz_api/internal/service/v1/routes/forecasts.py
braddf: I think the `null` is technically less accurate, leaving it out of the response was actually my "look at this later" but at least it's not confusing/wrong in the meantime

peterdudfield: ok, leave this in then. (one thing might be confusing, is sometimes it appears in the response, sometime it doesnt, but i guess the user will be actively setting this variable)

src/quartz_api/internal/service/v1/test_router.py
braddf: Done: https://github.com/openclimatefix/quartz-api/issues/314

peterdudfield (approved):

PR #321, opened 2026-06-01
Satellite UI API to S3 Presigned URL 

suvanbanerjee:
# Pull Request

## Description

Add endpoint to get s3 URL of a specific layer and channel 

## Checklist:


src/quartz_api/internal/s3.py
```diff
@@ -0,0 +1,48 @@
+"""S3 client."""
+import boto3
+from botocore.client import Config
+from botocore.exceptions import ClientError
```
peterdudfield: You might want to consider using `fsspec`, can be slightly easier to use that `boto3` and `botocore`. We have tended to use `fsspec` in some other mircoservices

src/quartz_api/internal/s3.py
```diff
@@ -0,0 +1,29 @@
+"""S3 client."""
+import fsspec
```
peterdudfield, line 4: looks good you are using fsspec now

peterdudfield (commented):
Thanks for using fsspec, I let @braddf review this properly, as I dont know the full ins and outs of this PR

src/quartz_api/internal/s3.py
braddf: Cool, fine with this for now but I think `boto3` is slightly more resilient in the pre-signed URL stuff. Can always tweak later if needed though!

src/quartz_api/internal/service/satellite/router.py
```diff
@@ -0,0 +1,55 @@
...
+    bucket = "historical-cloud-data-geotiff"
+    key = f"layers/{channel}/{timestamp.strftime('%Y%m%d_%H%M%S')}.tif"
+
+    if not s3_client.object_exists(bucket, key):
+        raise HTTPException(
+            status_code=404,
+            detail="File not found for the given channel and timestamp",
+        )
+
+    return HistoricSatelliteData(url=s3_client.get_presigned_url(bucket, key))
```
braddf, line 58: I think technically this pre-signed URL function should raise its own 404 error, so you could just wrap this in a try/catch and only make one request each time (all being well) instead of two?

src/quartz_api/internal/service/historic_satellite_data/router.py
```diff
@@ -0,0 +1,55 @@
...
+from typing import Annotated
+
+from fastapi import APIRouter, Depends, HTTPException
+
+from quartz_api.internal.s3 import S3Client, get_s3_client
+
+from .endpoint_types import HistoricSatelliteData
+
+router = APIRouter(
+    prefix="/historic-satellite-data",
```
braddf: I'd also simplify all these titles/routes to just `/satellite` if it were me, it's implicitly in the past for normal, non-Cloudcasting satellite data, and the "data" in the name is redundant

src/quartz_api/internal/service/historic_satellite_data/router.py
```diff
@@ -0,0 +1,55 @@
...
+    s3_client: S3ClientDep,
+) -> HistoricSatelliteData:
+    """Get a pre-signed URL for a historic satellite data file."""
+    if channel not in VALID_CHANNELS:
+        raise HTTPException(
+            status_code=400,
+            detail=f"Invalid channel. Must be one of {sorted(VALID_CHANNELS)}",
+        )
+
+    bucket = "historical-cloud-data-geotiff"
```
braddf: Could we set this in an ENV var, best to be able to change/set these things if needed without a code change

src/quartz_api/internal/s3.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+_s3_client: S3Client | None = None
+
+
+def get_s3_client() -> S3Client:
+    """Get the S3 client."""
+    global _s3_client
+    if _s3_client is None:
+        _s3_client = S3Client(region_name="eu-west-1")
```
braddf: Same here re ENV var (we can always have a default for these still though)

src/quartz_api/internal/service/satellite/router.py
```diff
@@ -0,0 +1,55 @@
...
+    "VIS006",
+    "VIS008",
+    "WV_062",
+    "WV_073",
+})
+
+S3ClientDep = Annotated[S3Client, Depends(get_s3_client)]
+
+
+@router.get("/", response_model=HistoricSatelliteData)
```
braddf, line 35: Need to add the usual auth guard on this route

src/quartz_api/internal/service/historic_satellite_data/__init__.py
```diff
@@ -0,0 +1,3 @@
+"""Historic satellite data router."""
+
+from .router import router
```
braddf: Need a blank line at end of file 🙏

src/quartz_api/internal/s3.py
braddf: And I think this will already be set by default potentially?

braddf (changes requested):
Looking good, pretty much there and think this might be slightly behind `main` now so needs branch updating

src/quartz_api/internal/service/satellite/router.py
suvanbanerjee: image

https://filesystem-spec.readthedocs.io/en/latest/api.html#fsspec.spec.AbstractFileSystem.sign


`ffspec` doesn't have an implementation for this hence the check is there so if I wrap it in a try catch and attempt to generate a pre-signed URL for a file that doesn't exist, it will still return a URL. However, the URL itself won't be valid, and opening it will result in an error because the underlying file is missing.

image

One possible option to remove the double call would be to shift the error handling to the frontend. The API would generate the URL, and on the frontend we would check whether it's valid or not. What do you think about this approach?

braddf: Ah in that case, keep it how it is for now 👍

braddf (approved):
Great stuff. We could have defaults in here for non-automatic ENV vars, but I think in this case if the bucket isn't found then we do want this to fail and know about it immediately – as long as it doesn't throw in a way that the API can't handle, this should be fine.

PR #323, opened 2026-06-05
Add icechunk to tif conversion endpoint and S3 upload support

suvanbanerjee:
# Pull Request

## Description

This PR adds the endpoint and logic to fetch files from Icechunk, convert them into TIFF format for UI display, and store the generated files in S3.

Chages needed in env

```
HISTORIC_SAT_S3_BUCKET=historical-cloud-data-geotiff
ICECHUNK_S3_BUCKET=nowcasting-sat-development
ICECHUNK_S3_PREFIX=odegree_v1/data/odegree_uk3000m.icechunk
AWS_DEFAULT_REGION=eu-west-1 <- this should be set by default on ec2 but on local we have to set it
ROUTERS=satellite
```

## Checklist:


src/quartz_api/internal/service/satellite/_ingest.py
```diff
@@ -0,0 +1,148 @@
...
+
+    dst_crs = CRS.from_epsg(4326)
+    dst_tf, dst_w, dst_h = calculate_default_transform(
+        src_crs, dst_crs,
+        ds["data"].shape[-1], ds["data"].shape[-2],
+        left=x.min() - xres / 2, bottom=y.min() - yres / 2,
+        right=x.max() + xres / 2, top=y.max() + yres / 2,
+    )
+
+    # Filter to last 48 hours
```
braddf: Update comment

src/quartz_api/internal/service/satellite/router.py
```diff
@@ -56,3 +58,19 @@ def get_historic_satellite_data_url(
...
+@router.post(
+    "/ingest",
+    status_code=status.HTTP_202_ACCEPTED,
+)
+def trigger_ingest(
+    background_tasks: BackgroundTasks,
+    _: AuthDependency,
+) -> IngestResponse:
+    """Trigger ingest of latest satellite data for all channels."""
+    background_tasks.add_task(run_ingest)
```
braddf, line 81: Doesn't look like this has any guarding against concurrent processing? This can be in-memory, check out the "pre-warm" caching function on the v1 for a similar thing; basically just set a flag to show if the processing is already active, and if it is don't kick off another task. 
This is MVP, we can be smarter about retry logic, queuing etc. later on, but let's just make sure this doesn't accidentally cause a bottleneck in the API.

src/quartz_api/internal/service/satellite/router.py
```diff
@@ -56,3 +58,19 @@ def get_historic_satellite_data_url(
         )
 
     return HistoricSatelliteData(url=s3_client.get_presigned_url(bucket, key))
+
+
+@router.post(
+    "/ingest",
+    status_code=status.HTTP_202_ACCEPTED,
+)
```
braddf, line 71: Also needs auth on this, probably with the admin role similar to the pre-warm tasks in v0/v1 UK router

src/quartz_api/internal/service/satellite/_ingest.py
```diff
@@ -0,0 +1,148 @@
...
+        Tuple of (latest timestamp, ts_str).
+    """
+    log.info("Ingest started")
+
+    s3_bucket = os.environ["HISTORIC_SAT_S3_BUCKET"]
+    icechunk_bucket = os.environ["ICECHUNK_S3_BUCKET"]
+    icechunk_prefix = os.environ["ICECHUNK_S3_PREFIX"]
+    region = os.environ.get("AWS_DEFAULT_REGION")
+
+    fs = s3fs.S3FileSystem(**({"client_kwargs": {"region_name": region}} if region else {}))
```
braddf: This should be using the S3Client abstraction, no? Instead of creating its own instance here, this might be duplicated across the two PRs – make sure there's only one implementation and use it everywhere

src/quartz_api/internal/service/satellite/_ingest.py
```diff
@@ -0,0 +1,148 @@
...
+
+            except Exception as e:
+                failed += 1
+                log.exception("Failed %s @ %s: %s", channel, ts_str, e)
+
+    latest_t = all_times[-1]
+    latest_ts = str(latest_t)[:19].replace("-", "").replace("T", "_").replace(":", "")
+    log.info(
+        "Ingest complete: %d uploaded, %d skipped, %d failed (latest=%s)",
+        uploaded, skipped, failed, latest_ts,
```
braddf, line 208: We're logging the failures here, but we might also want to send these to Sentry too if there are any, depending how often they occur... Add in a `capture_message` (I think that's the method) so we have visibility, and we can always mute or remove later if it's noisy.

src/quartz_api/internal/service/satellite/_ingest.py
```diff
@@ -0,0 +1,148 @@
...
+
+    Uploads any missing channel+timestamp combos from the last 48 hours,
+    skipping ones already present in S3.
+
+    Returns:
+        Tuple of (latest timestamp, ts_str).
+    """
+    log.info("Ingest started")
+
+    s3_bucket = os.environ["HISTORIC_SAT_S3_BUCKET"]
```
braddf: Remember to update these with changes from #321 when you get that working 👍

src/quartz_api/internal/service/satellite/_ingest.py
```diff
@@ -0,0 +1,148 @@
...
+import rasterio
+import s3fs
+import xarray as xr
+from rasterio.crs import CRS
+from rasterio.warp import Resampling, calculate_default_transform, reproject
+from rasterio.windows import from_bounds as window_from_bounds
+
+log = logging.getLogger(__name__)
+
+LEFT, BOTTOM, RIGHT, TOP = -20.0, 45.8, 15.0, 64.2
```
braddf: Short comment stating what these are for would help future colleagues get up to speed 😊
(I remember these from GSoC and our conversations, but otherwise it's a guessing game!)

braddf (changes requested):
Looking good, some small comments to address once #321 is finalised

suvanbanerjee:
Regarding commit https://github.com/openclimatefix/quartz-api/pull/323/commits/20a54eb9cde99293c490eb8df1824ab3203833ca in PR #323.

The conversion functions were producing visible artifacts during nighttime imagery. the artifacts were caused by sensor noise in low-signal conditions rather than the reprojection itself. I tested multiple reprojection methods, but all of them still produced similar artifacts.

Nighttime artifacts

My first approach was to filter the noise using a fixed threshold. However, this did not generalize across channels because the noise floor varies significantly.

For example:

- `VIS006` could be filtered effectively with a threshold around `4.0-5.0`.
- `IR_016` required thresholds around `10.0`.

As a result, a single static threshold was not suitable.

Static threshold comparison

The next approach was to use a dynamic threshold based on image percentiles rather than a fixed value.

After testing different percentile values, the `99th percentile (P99)` consistently produced the best results for the affected channels.

### VIS006 percentile sweep

local_cliff_sweep_VIS006

### VIS008 percentile sweep

local_cliff_sweep_VIS008

While P99 worked well at night, it introduced a new issue during daytime imagery. Daytime P99 values can become very large (often >200), which resulted in valid cloud structures being removed.

To address this, I added a channel-specific cliff limit that caps the threshold value. The cliff limits were selected based on analysis of approximately one month of historical data:

| Channel | Peak Max | Avg Mean | Avg Mode | Peak P95 | Peak P99 | Peak P99.9 |
|----------|----------:|----------:|----------:|----------:|----------:|------------:|
| IR_016 | 40.031 | 6.544 | 5.395 | 8.281 | 9.922 | 24.865 |
| VIS006 | 16.688 | 2.265 | 2.294 | 2.715 | 3.857 | 9.711 |
| VIS008 | 33.250 | 4.471 | 4.457 | 5.102 | 7.570 | 20.000 |

The final configuration is:

```python
SOLAR_NOISE_CONFIG = {
    "IR_016": {"percentile": 99.0, "cliff_limit": 12.0},
    "VIS006": {"percentile": 99.0, "cliff_limit": 5.0},
    "VIS008": {"percentile": 99.0, "cliff_limit": 10.0},
}
```

This removes the nighttime sensor noise while preserving cloud structures during daytime imagery.

braddf (approved):
Nice. Let's get this merged and see how it does!

PR #326, opened 2026-06-15
Update resp fields and implement put_actual_generation & put_location methods

katyalmohit:
# Pull Request

## Description

- Updated model returns (`power_kw` to `PowerKW` and `time` to `Time`)
- Implemented `put_actual_generation` method
- Implemented `put_location` method
- Added test

Fixes
- https://github.com/openclimatefix/client-private/issues/475
- https://github.com/openclimatefix/client-private/issues/476
- https://github.com/openclimatefix/client-private/issues/477

## How Has This Been Tested?

- [x] Integration test
- [x] CI

## Checklist:

### DIFF
--- a/src/quartz_api/internal/backends/dataplatform/client.py
+++ b/src/quartz_api/internal/backends/dataplatform/client.py
@@ -517,9 +517,23 @@ class StorageClient(models.StorageInterface):
             location_uuid=location.uuid,
         )
         if not existing:
-            raise HTTPException(
-                status_code=404,
-                detail=f"No location found for UUID '{location.uuid}'",
+            create_req = messages_pb2.CreateLocationRequest(
+                location_name=location.name,
+                energy_source=energy_type_map[energy_type],
+                geometry_wkt=f"POINT ({location.longitude} {location.latitude})",
+                effective_capacity_watts=int(location.capacity_kilowatts * 1000),
+                location_type=location_type_map[location_type],
+                metadata=dict_to_struct(location.metadata),
+            )
+            created = await self.dpc.CreateLocation(create_req)
+            return models.Location(
+                uuid=UUID(created.location_uuid),
+                name=created.location_name,
+                latitude=location.latitude,
+                longitude=location.longitude,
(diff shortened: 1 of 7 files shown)


src/quartz_api/internal/backends/dataplatform/client.py
```diff
@@ -479,8 +510,43 @@ async def put_location(
...
+            location_type=location_type,
+            authdata=authdata,
+            location_uuid=location.uuid,
+        )
+        if not existing:
+            raise HTTPException(
+                status_code=404,
+                detail=f"No location found for UUID '{location.uuid}'",
+            )
+        current = existing[0]
```
devsjc, line 538: So we can't create locations with this method, is that right?

devsjc: Also I don't know if this needs some awareness of ownership? E.g. using the UpdateLocationOwner RPC in case it's a sites user making a new location?

src/quartz_api/internal/backends/dataplatform/test_client.py
devsjc, line 1: Nice, thank you for mocking

devsjc (approved):

src/quartz_api/internal/backends/dataplatform/client.py
katyalmohit: > So we can't create locations with this method, is that right?

Good catch, you're right. I've now updated it - so now it creates when the location doesn't exist

katyalmohit: > Also I don't know if this needs some awareness of ownership? E.g. using the UpdateLocationOwner RPC in case it's a sites user making a new location?

It's using version `0.29.0` of data-platform and `UpdateLocationOwner` only came in `0.32.0`. Shall I upgrade it or maybe in a different PR?

devsjc: I think a different PR is fine!

devsjc (approved):

PR #330, opened 2026-06-19
fix: pass auth_instance to sentry middleware

### LINKED ISSUE #329: add user name to sentry logs
It might be useful to add emails to sentry logs.

This could be done with 
https://docs.sentry.io/platforms/python/enriching-events/tags/ and could be done in the auth middleware

suvanbanerjee:
# Pull Request

## Description

Sentry was only showing client IP addresses and was not logging user email information because the auth instance was not being passed through.

Fixes #329 

## Checklist:

### DIFF
--- a/src/quartz_api/internal/middleware/sentry.py
+++ b/src/quartz_api/internal/middleware/sentry.py
@@ -8,6 +8,7 @@ from fastapi import FastAPI, Request, Response
 from starlette.middleware.base import BaseHTTPMiddleware
 
 from quartz_api.internal.middleware import auth
+from quartz_api.internal.middleware.auth import get_org_id_from_authdata
 
 
 class SentryUserMiddleware(BaseHTTPMiddleware):
@@ -41,6 +42,9 @@ class SentryUserMiddleware(BaseHTTPMiddleware):
                             "email": payload.get(auth.EMAIL_KEY),
                         },
                     )
+                    org_id = get_org_id_from_authdata(payload)
+                    if org_id:
+                        sentry_sdk.set_tag("hubspot_company_id", org_id)
             except Exception:
                 # silently fail to not break requests
                 logging.debug("Could not extract user for Sentry")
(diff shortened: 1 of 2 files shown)


suvanbanerjee:
@braddf I decoded the current JWT and these are the fields present:

```
{
  "trial_ends_at": "...",
  "https://openclimatefix.org/email": "...",
  "iss": "https://nowcasting-pro.eu.auth0.com/",
  "sub": "google-oauth2|10...34",
  "aud": [
    "https://api.nowcasting.io/",
    "https://nowcasting-pro.eu.auth0.com/userinfo"
  ],
  "iat": 1781842439,
  "exp": 1782706439,
  "scope": "openid profile email offline_access",
  "azp": "...",
  "permissions": [
    "read:trial"
  ]
}
```

When we add the organization_id, what key should we use to store it in the JWT?

I can also add it to the Sentry middleware so it gets attached to Sentry events.

src/quartz_api/internal/middleware/sentry.py
```diff
@@ -41,6 +42,9 @@ async def dispatch(
                             "email": payload.get(auth.EMAIL_KEY),
                         },
                     )
+                    org_id = get_org_id_from_authdata(payload)
```
peterdudfield, line 45: Don't worry about organization I'd, just ignore that

braddf:
@suvanbanerjee the org id should be stored/fetched already here:

https://github.com/openclimatefix/quartz-api/blob/main/src/quartz_api/internal/middleware/auth.py#L120

So you can use that key too

src/quartz_api/internal/middleware/sentry.py
peterdudfield: if the email is there, thats fine

suvanbanerjee:
> @suvanbanerjee the org id should be stored/fetched already here:
> 
> https://github.com/openclimatefix/quartz-api/blob/main/src/quartz_api/internal/middleware/auth.py#L120
> 
> So you can use that key too

I used the `get_org_id_from_authdata` function rather than accessing `app_metadata["hubspot_id"]` directly. That way, if the underlying key ever changes in the future, we only need to update it in one place instead of everywhere it's used.

braddf (approved):

PR #333, opened 2026-06-24
fix: restrict organisation ID filtering to SITE/SUBSTATION location types in Data Platform client

PavanRaghavendraKulkarni:
## Description

Fixes the Data Platform client so that organisation ID filtering is only applied to `SITE` and `SUBSTATION` location types, not to regional or national queries.

## Motivation

Prior to this change, the organisation ID filter was applied to **all** location types when an auth token was present. This inadvertently blocked Adani site forecast and generation queries because the org filter was also being applied inside `get_predicted_generation` and `get_actual_generation` for non-site location types.

## Changes

**`src/quartz_api/internal/backends/dataplatform/client.py`**
- `get_predicted_generation`: org ID filter now only applied when `location_type` is `SITE` or `SUBSTATION`
- `get_actual_generation`: `_check_user_access` now only called for `SITE` location type
- `put_actual_generation`: `_check_user_access` now only called for `SITE` location type

## Testing

Manually verified against local dev environment with a valid auth token:

| Endpoint |
|---|
| `GET /sites` 
| `GET /sites/{uuid}/forecast` (org site) 
| `GET /sites/{uuid}/generation` (org site) 
| `GET /sites/{uuid}/forecast` (other site)


src/quartz_api/internal/service/regions/router.py
```diff
@@ -25,6 +25,33 @@
 )
 
 router = APIRouter(tags=[pathlib.Path(__file__).parent.stem.capitalize()])
+SITE_OBSERVER_NAME = "site_api"
```
devsjc: The observer is called "site_api"? Does this mean every user writes their data to the same observer? I wonder if we're not better off with an observer per user...

PavanRaghavendraKulkarni: Yes, I discussed this with @peterdudfield . One of the main challenges was determining the observer name for a location when the user comes through the API endpoint. As a solution, we set it to `site-api`.

For debugging and audit purposes, we always have the auth data available, so we can determine who accessed the endpoint or performed the write operation based on that information.

That said, your suggestion makes sense, and I'm open to discussing a better approach if we think there's a more suitable solution.

peterdudfield: I think for adani for this route, this should keep it simple

devsjc: If we're talking about ad sites; shouldn't the logic for that be in the `sites` router? This is the regions router...?

PavanRaghavendraKulkarni: The value "site_api" is used here because Indian regional actuals (e.g. RUVNL) were historically uploaded via the Site API, so the Data Platform stores them under that observer name

PavanRaghavendraKulkarni: I can change it to REGIONAL_OBSERVER_NAME if naming convention is the issue

peterdudfield: You are right @devsjc . This is the regions/router.py. This is nothing to do with Adani, and we dont need any filter of org id in this router.py.

peterdudfield: If this is for ruvnl, I thought the observer was defined here https://github.com/openclimatefix/solar-consumer/blob/main/solar_consumer/save/save_data_platform.py#L60

peterdudfield (commented):
Im not sure the title of this PR is correct

src/quartz_api/internal/backends/dataplatform/client.py
```diff
@@ -96,7 +96,15 @@ async def get_predicted_generation(
                 minutes=forecast_horizon_minutes,
             )
 
-        organisation_id: str | None = get_org_id_from_authdata(authdata) if authdata != {} else None
+        _site_level = location_type in (
+            models.LocationType.SITE,
+            models.LocationType.SUBSTATION,
```
peterdudfield, line 101: should remove substation?

peterdudfield:
Im not sure this is needed. For Adani there are only sites.

PR #334, opened 2026-06-29
chore: remove db audit logging as it is handled by apitally

suvanbanerjee:
# Pull Request

## Description

Remove the RequestLoggerMiddleware (db audit logging), as request logging is now handled by Apitally. This eliminates the Failed to log request to DB: Database client not found in app state Sentry events which were the dominant source of Sentry quota consumption (~101K out of 140K total requests in the last 90 days), mostly triggered by automated scanners hitting 404 endpoints.

## How Has This Been Tested?

Tested locally by removing the audit middleware and sending a 404 request, confirmed no sentry error via local logs.

- [x] Yes

## Checklist:


peterdudfield (approved):

peterdudfield (changes requested):
wait

src/quartz_api/cmd/main.py
```diff
@@ -464,8 +464,6 @@ def redoc_html() -> FileResponse:
         allow_methods=["*"],
         allow_headers=["*"],
     )
-    if conf.get_string("backend.source") != "dataplatform":
```
peterdudfield, line 466: so this doesnt run for the dataplatform background.

What services are we getting this error for?

peterdudfield: can you see what API we are getting this warning from

suvanbanerjee: looks like they are hitting out server ip directly  and not api.quartz.solar 

image

`uk-api` on sentry has around 52k events in devleopemnt and 49.5k events on prod,

peterdudfield: and do you know which api it is?

suvanbanerjee: @peterdudfield I'm not entirely sure which environment is firing these, here's what I found:

I found this request in Sentry hitting `http://54.155.28.70/`:
sentry request

When I visited `http://54.155.28.70/docs` it only serves substations endpoints, which points to `uk-development-quartz-api` with `ROUTERS=substations`:
substations docs

However when I checked the AWS config for `uk-development-quartz-api`, it has `SOURCE=dataplatform`, which means the audit middleware guard skips it entirely:
source config

So I'm not sure how it's still firing, do you have any idea what I'm missing here?

suvanbanerjee: on thing, not sure might be the cause, that the API  version running on EBS is currently on 0.3.27 (fetched from EBS logs and checked on teraform)

image

and this dataplatfom exclusion code was added in d0f7eb1e9b212efa1d2ab2bc29ec18a1d683b3d1 on 27may (around a month back)

image

peterdudfield: ah ok, so we need to upgrade the substation api, from 0.3.27 to the latest. Would that solve it?

suvanbanerjee: Yes it should

PR #335, opened 2026-07-01
feat: fix api migrations

PavanRaghavendraKulkarni:
# Pull Request

## Description

This PR brings the sites API responses and structures in line with the old API specs for migration compatibility. It also cleans up model-level schema metadata configurations, adds automatic model resolution support for solar/wind forecasters, and cleans up fields/naming across the sites module.

### Key Changes:
1. **Sites API Response Compatibility**:
   * Changed `capacity_kW` to `capacity_kw` across the API (`SiteProperties`, `Site`, routes, and test suite).
   * Removed the `metadata` dictionary field from the sites responses (`SiteProperties` and `Site`) as it was not present in the old API.
   * Stripped out the extra fields (`initialization_timestamp_utc` and `plevel_kW`) from the site `PredictedPower` model and `location_uuid` from `ActualPower` to mirror the old API specs exactly.
   * Added field validators to round presentation coordinate floats (`latitude` and `longitude`) to 4 decimal places.
2. **Forecaster Resolution**:
   * Integrated `_resolve_forecaster_name` helper to dynamically choose wind/solar forecaster models (e.g. `windnet_ad_sites`, `pvnet_extra_ad_sites_ecmwf_satellite_pv_long`) using metadata overrides or default fallbacks based on site energy types, which is then passed to the Data Platform query.
3. **Pydantic Model Schema Improvements**:
   * Removed redundant `json_schema_extra` parameter nesting on model fields in favor of direct `description` parameters on Pydantic's `Field` objects.

Fixes # (specify issue if applicable)

## How Has This Been Tested?

The changes were validated using the local test suite:
* Run the unit test suite:
  ```bash
  uv run pytest -k "not test_quartzdb and not test_app"

## Checklist:


src/quartz_api/internal/service/sites/router.py
```diff
@@ -137,23 +176,28 @@ async def get_forecast(
 ) -> list[PredictedPower]:
     """Get forecast of a site (Solar or Wind, auto-detected)."""
     site = await _get_site_with_energy_type(db, site_uuid, auth)
+
```
PavanRaghavendraKulkarni, line 179: Hey @peterdudfield, could you please review this? It works for Adani. However, if we consider it from a more generic perspective, it should also work because the models will differ between the UK and India.

src/quartz_api/internal/service/sites/router.py
```diff
@@ -61,13 +102,12 @@ async def get_sites(
     out: list[Site] = [
         Site(
             site_uuid=s.uuid,
-            client_site_name=s.name,
+            client_site_name=s.metadata.get("client_site_name") or s.name,
```
PavanRaghavendraKulkarni, line 105: If the metadata doesn't contain `client_site_name`, the location name will be used instead.

src/quartz_api/internal/service/sites/router.py
```diff
@@ -13,12 +14,52 @@
...
+            site.uuid,
+            name,
+        )
+        return name
+
+    raise ValueError(
+        f"Site {site.uuid} has no forecast_name in metadata and an unrecognised "
+        f"energy_type '{energy_type}'. Set metadata.forecast_name explicitly.",
+    )
+
```
PavanRaghavendraKulkarni, line 62: These changes are very specific to India (Adani). We should come up with a more generic solution that works across different regions and models.

src/quartz_api/internal/service/sites/endpoint_types.py
```diff
@@ -1,58 +1,91 @@
...
 
 class Site(SiteProperties):
     """Site information, including properties and unique identifier."""
 
     site_uuid: UUID = Field(
         ...,
-        json_schema_extra={"description": "The unique identifier for the site."},
+        description="The site uuid assigned by ocf.",
+    )
+    latitude: float = Field(
```
peterdudfield: I think these are got from SiteProperties

peterdudfield: so you dont need these

src/quartz_api/internal/service/sites/router.py
peterdudfield: yep, in the future

src/quartz_api/internal/service/sites/router.py
peterdudfield: yep this is only for Adani, for UK we want to make a new `v1/sites`, but lets get this out there and working first

peterdudfield (changes requested):

src/quartz_api/internal/service/sites/endpoint_types.py
PavanRaghavendraKulkarni: resolved

peterdudfield (approved):

PR #341, opened 2026-07-08
feat: give full access to ocf admin

PavanRaghavendraKulkarni:
# Pull Request

## Description

This pull request updates the authentication middleware to bypass organization-based filtering when the authenticated user has OCF Admin privileges (`ocf:admin` permission). 

Previously, all site-related queries were filtered by the user's HubSpot company ID (`organisation_id_filter`), preventing OCF admins from viewing or managing sites across multiple organizations. With this change, `get_org_id_from_authdata` returns `None` for admin users, enabling unfiltered access to all locations.

### Key Changes
1. **Middleware Update**: Added an admin check in `get_org_id_from_authdata` (in `src/quartz_api/internal/middleware/auth.py`).
2. **Unit Tests**: Created a new test file `src/quartz_api/internal/middleware/test_auth.py` to cover:
   - Extraction of HubSpot company ID for regular users.
   - Bypassing the company ID and returning `None` for admin users.
   - Handling tokens without metadata.

[PR](https://github.com/openclimatefix/client-private/issues/545)
## How Has This Been Tested?

- Run locally using `uv run pytest src/quartz_api/internal/middleware/test_auth.py` (all tests passed).
- Verified manually using curl request testing.

- [x] Yes

## Checklist:

### DIFF
--- a/src/quartz_api/internal/backends/dataplatform/client.py
+++ b/src/quartz_api/internal/backends/dataplatform/client.py
@@ -98,6 +98,11 @@ class StorageClient(models.StorageInterface):
             )
 
         organisation_id: str | None = get_org_id_from_authdata(authdata) if authdata != {} else None
+        if organisation_id == "no-org-access" and location_type == models.LocationType.SITE:
+            raise HTTPException(
+                status_code=403,
+                detail="User is not associated with any organization.",
+            )
         req = messages_pb2.ListLocationsRequest(
             location_uuids_filter=[str(location_uuid)],
             energy_source_filter=energy_type_map[energy_type],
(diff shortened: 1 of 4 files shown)


src/quartz_api/internal/backends/dataplatform/client.py
```diff
@@ -98,6 +98,11 @@ async def get_predicted_generation(
             )
 
         organisation_id: str | None = get_org_id_from_authdata(authdata) if authdata != {} else None
+        if organisation_id == "no-org-access" and location_type == models.LocationType.SITE:
```
katyalmohit: Looks good! Just one thing I want to check - here we only return a 403 for sites, but for a no-org user we still pass "no-org-access" as the org filter for regions too. Since regions don't have an org, wouldn't that come back empty and give a 404? (Just trying to understand if we only want to protect sites here. If so, maybe we set the org id back to None for non-sites.)

katyalmohit (approved):

PR #346, opened 2026-07-14
feat: blackout and latest flag sat get api

suvanbanerjee:
# Pull Request

## Description

Add a blackout period, as the percentile filters are not accurate enough and the image is completely black at night for certain channels.

Add the latest flag to the GET sat route, to find the latest available timestamp to show on live in FE

## How Has This Been Tested?

- [x] Locally

## Checklist:

### DIFF
--- a/src/quartz_api/internal/s3.py
+++ b/src/quartz_api/internal/s3.py
@@ -1,4 +1,6 @@
 """S3 client."""
+from datetime import UTC, datetime, timedelta
+
 import fsspec
 
 # Satellite S3 config, populated at startup via configure() from the parsed
@@ -81,6 +83,19 @@ class S3Client:
     def object_exists(self, bucket: str, key: str) -> bool:
         """Check if an object exists in S3."""
         return self.fs.exists(f"s3://{bucket}/{key}")
+
+    def get_latest_key(
+        self, bucket: str, prefix: str, lookback_minutes: int = 30,
+    ) -> str | None:
+        """Probe 5-min-cadence keys backward from now, return the first one that exists."""
+        ts = datetime.now(tz=UTC).replace(second=0, microsecond=0)
+        ts -= timedelta(minutes=ts.minute % 5)
+        for _ in range(lookback_minutes // 5 + 1):
+            key = f"{prefix}{ts:%Y%m%d_%H%M%S}.tif"
+            if self.object_exists(bucket, key):
+                return key
+            ts -= timedelta(minutes=5)
+        return None
 
     def upload_bytes(self, bucket: str, key: str, data: bytes) -> None:
         """Upload raw bytes to an S3 key."""
--- a/src/quartz_api/internal/s3.py
+++ b/src/quartz_api/internal/s3.py
@@ -87,7 +87,11 @@ class S3Client:
     def get_latest_key(
         self, bucket: str, prefix: str, lookback_minutes: int = 30,
     ) -> str | None:
-        """Probe 5-min-cadence keys backward from now, return the first one that exists."""
+        """Search backwards in 5-minute intervals and return the first available key.
+
+        Used to show the LIVE cloud image in the UI. The default 30-minute lookback
+        keeps the image recent while remaining visually distinct.
+        """
         ts = datetime.now(tz=UTC).replace(second=0, microsecond=0)
         ts -= timedelta(minutes=ts.minute % 5)
         for _ in range(lookback_minutes // 5 + 1):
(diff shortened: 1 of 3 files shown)


src/quartz_api/internal/service/satellite/_ingest.py
```diff
@@ -148,12 +141,14 @@ def _run_ingest() -> tuple[str, str]:
...
-                if channel in LAYER_CONFIG:
-                    config = LAYER_CONFIG[channel]
+                if ts_str[9:13] in {t.replace(":", "") for t in config.get("blackout", [])}:
+                    dst_arr = np.zeros((dst_h, dst_w), dtype=np.float32)
+                else:
+                    chan_idx = list(ds.channel.values).index(channel)
+                    arr = ds["data"].isel(time=t_idx, channel=chan_idx).values.astype(np.float32)
+                    arr = np.flipud(arr)
 
                     if "percentile" in config:
```
suvanbanerjee: I have not remove this, as i think it might be useful in future, maybe later i think i can separate this in a new file to make the code more maintainable

src/quartz_api/internal/s3.py
```diff
@@ -82,6 +84,19 @@ def object_exists(self, bucket: str, key: str) -> bool:
         """Check if an object exists in S3."""
         return self.fs.exists(f"s3://{bucket}/{key}")
 
+    def get_latest_key(
+        self, bucket: str, prefix: str, lookback_minutes: int = 30,
```
suvanbanerjee, line 88: `lookback_minutes` is the time it will look back to, 30min will prove a visually distinct image and will still be good for the `LIVE` timestamp

src/quartz_api/internal/service/satellite/_ingest.py
```diff
@@ -24,11 +24,11 @@
 LEFT, BOTTOM, RIGHT, TOP = -17.05, 46.49, 11.60, 63.31
 # How far back to backfill missing data (in hours)
 BACKFILL_HOURS = 48
-# config to remove artifacts during nighttime and invert channels
+# Per-channel nighttime cleanup, inversion, and recurring "HH:MM" (UTC) blackout times.
 LAYER_CONFIG = {
-    "VIS006": {"percentile": 99.0, "cliff_limit": 5.0},
-    "VIS008": {"percentile": 99.0, "cliff_limit": 10.0},
-    "IR_016": {"percentile": 99.0, "cliff_limit": 12.0},
+    "VIS006": {"blackout": ["23:45", "00:15", "00:45", "01:15"]},
```
braddf, line 29: Interesting, so does it not get the same artefacts in the early morning hours? I guess maybe because the angle of incidence of the satellite 🤔

src/quartz_api/internal/service/satellite/_ingest.py
braddf: What exactly does this do / will it do again in future..?

src/quartz_api/internal/s3.py
braddf: Maybe add this in a comment in the code if not already, helps those new to the code understand the roles of variables like this 😊

src/quartz_api/internal/service/satellite/router.py
```diff
@@ -42,18 +42,35 @@
...
     bucket = get_geotiff_bucket()
+
+    if latest:
+        key = s3_client.get_latest_key(bucket, f"layers/{channel}/")
+        if key is None:
+            raise HTTPException(
+                status_code=404,
+                detail="No files found for the given channel in the last 30 minutes",
+            )
+        return HistoricSatelliteData(url=s3_client.get_presigned_url(bucket, key))
```
braddf, line 66: And do we know from the response what timestamp was selected, or not? Could be useful to know for UI in future

braddf (changes requested):
Couple of comments, otherwise good!

src/quartz_api/internal/service/satellite/router.py
suvanbanerjee: Not directly in the response, We can determine it easily. The returned filename contains the timestamp, so I don't think we need to include it here.

src/quartz_api/internal/service/satellite/_ingest.py
suvanbanerjee: No, it only occurs at those specific timestamps. Looking into it, I think it's plausibly related to the angle of incidence, or possibly [sunglint](https://en.wikipedia.org/wiki/Sunglint). I think blacking out those periods is the most reliable fix, as filtering the noise still leaves some room for error.

src/quartz_api/internal/service/satellite/_ingest.py
suvanbanerjee: The original intent was noise filtering, but since that didn't work reliably, I kept it in case we could repurpose it for land masking. I've just removed it from the PR. and if need i might ref old commits

src/quartz_api/internal/service/satellite/router.py
braddf: Great, I was hoping that was the case – sounds good 👍

braddf (approved):

PR #347, opened 2026-07-15
feat: implement Quartz DB client and integrate rate limiting middleware

PavanRaghavendraKulkarni:
## Description
Fixes rate limiting not being enforced on `/v1` routes in the development environment, despite working correctly on local.

**Root cause:** `SlowAPIMiddleware` relies on walking `app.routes` and matching each route's `.endpoint` attribute to apply `default_limits`. In the FastAPI version running on dev, included routers are wrapped internally as `_IncludedRouter` objects, which don't expose `.endpoint` the way slowapi expects. This caused slowapi's route resolution to silently fail and exempt all `/v1/*` requests from rate limiting — even though the exact same code correctly rate-limited requests locally.

**Fix:** Moved from relying on `default_limits` + middleware auto-discovery to explicit `@limiter.limit("20/second;3600/hour")` decorators on every v1 endpoint (16 total, across `discovery.py`, `regions.py`, `forecasts.py`, `generation.py`). Decorators attach the limit directly to the endpoint function, so enforcement no longer depends on slowapi correctly discovering the route via `app.routes`.

Fixes #

## How Has This Been Tested?
- Confirmed locally that rate limiting worked before this change (default_limits + middleware sufficed).
- Reproduced the dev-only bypass: blasted concurrent requests at `/v1/sources` on dev — all returned `200`, no `429`, confirming middleware-based auto-discovery was failing there.
- Deployed the decorator-based fix to dev and re-ran the same load test: 45 parallel requests to `/v1/sources` → Total 

requests: 30
✅ 200 OK: 24
🚫 429 Too Many Requests: 6

- [x] Yes

## Checklist:

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -9,7 +9,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     git
 
 WORKDIR /opt/app
-COPY pyproject.toml /opt/app/pyproject.toml
+COPY pyproject.toml uv.lock /opt/app/
 
 # * Compile bytecode to reduce startup time
 # * Disable cache to reduce image size
@@ -18,7 +18,7 @@ ENV UV_COMPILE_BYTECODE=1 \
     UV_LINK_MODE=copy
 
 RUN mkdir src && \
-    uv sync --no-dev --no-install-project --no-editable
+    uv sync --locked --no-dev --no-install-project --no-editable
 
 RUN rm -rf /opt/app/.venv/lib/python3.12/site-packages/**/tests
 
@@ -30,7 +30,7 @@ FROM build-deps AS build-app
 COPY src /opt/app/src
 COPY .git /opt/app/.git
 COPY README.md /opt/app/README.md
-RUN uv sync --no-dev --no-editable
+RUN uv sync --locked --no-dev --no-editable
 
 # Delete package tests
 RUN rm -rf /opt/app/.venv/lib/python3.12/site-packages/**/test_*
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,7 +20,7 @@ classifiers = ["Programming Language :: Python :: 3"]
 dependencies = [
     "cryptography >= 41.0.0",
     "dp-sdk >= 0.32.0,<1.0.0",
-    "fastapi>=0.105.0",
+    "fastapi==0.136.3", # 0.137.0 currently has a bug with ratelimits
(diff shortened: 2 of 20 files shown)


braddf:
@PavanRaghavendraKulkarni – I had a dig into this weirdness between local and dev, and the need for explicit decorators on all routes – it seems that in the newer versions of FastAPI + `slowapi` in combination, there's a slight mismatch and the "old way" of setting the default/global rate limits aren't actually being picked up. I'm fine with them being set on each route for now to ensure they are active, which seems to be the case now from testing, and if this is fixed in a future version we can revisit. 

I reckon the reason that this was hard to debug between local and dev was that the uv lockfile wasn't being copied and enforced on the container build, so the versions we would run locally weren't guaranteed to be the same as on dev. I've made sure this is now copied in the Containerfile so it should now be fixed to that and use the same versions as in local.

N.B. I've also added in v0 routes, as it appears that this applies across all routes, so with the newer package versions installed, this is also an issue.

braddf:
Also, coincidentally, I had a hard time working out if the rate limiting was working using Postman, as when using a bash command to test ~200 parallelised requests to `/v1/sources` it was getting 429s. Turns out a 429 code isn't classed as an "error" in Postman, you have to explicitly check what code is being returned in the Postman Runner 😂  lesson learned! Bad assumptions hide bugs haha.

braddf (approved):

braddf:
@peterdudfield I've reviewed but also added bits, so double check when you're back!

peterdudfield:
wow good find @braddf! im interested how you managed to find this?

How many versions a go do we need to pin back due to this bug of `fastapi`? It might be neater just to do that, and write a bug report on `slowapi` and see if they can fix it. 

If not i'd probably do a import for `"20/second;3600/hour"` rather than writing it out everytime.

Otherwise I think looks good

peterdudfield:
as discussed with Software team in retro. We should pin fastapi back to the version we know it was working, with a comment about why. In the future that bug should be fixed and we can remove that pin. @PavanRaghavendraKulkarni could you do this (if you dont know which version to pin back to, then please ask @braddf )

Something that isnt clear to me right now, is why the test didnt catch this?

peterdudfield:
I found that the `fastaspi==0.137.0` fails the rate limiting test, but `0.136.3` is fine. This is being pinned in #349

peterdudfield:
I think we can actually close this, as its all works in #349 now

PR #356, opened 2026-08-03
fix: comment out inversion for IR_039 in layer configuration

suvanbanerjee:
# Pull Request

## Description
comment out inversion for IR_039 in layer configuration

## Checklist:


braddf (changes requested):
Probably remove / empty dict for this rather than commenting out

braddf (approved):
As long as that's fine not being in there! Might be cleaner to retain all of the layers in that config enum but up to you 👍

PR #368, opened 2026-08-17
feat: add composite sat channel processing on backend

### LINKED ISSUE #367: Satellite UI: Move the composite layers to backend
Move the layer composition logic to the backend. This will improve frontend performance and reduce the processing load on the client browser.

suvanbanerjee:
# Pull Request

## Description

Move the layer composition logic to the backend

New files are stored in the same S3 with the composite layer config named defined as

```
COMPOSITE_CONFIG: dict[str, list[str]] = {
    "COMPOSITE_VISIBLE": ["IR_016", "VIS008", "VIS006"],
    "COMPOSITE_INFRARED": ["IR_134", "IR_097", "IR_120", "IR_087", "IR_108"],
    "COMPOSITE_BLUE": ["WV_073", "WV_062"],
}
``` 

Fixes #367 

## How Has This Been Tested?

- [x] Locally then plotted on [geotiff.io](https://geotiff.io/)

image
image
image

## Checklist:


src/quartz_api/internal/service/satellite/_ingest.py
```diff
@@ -42,6 +44,73 @@
...
+    "COMPOSITE_VISIBLE": ["IR_016", "VIS008", "VIS006"],
+    "COMPOSITE_INFRARED": ["IR_134", "IR_097", "IR_120", "IR_087", "IR_108"],
+    "COMPOSITE_BLUE": ["WV_073", "WV_062"],
+}
+
+SAT_MAX_ALPHA = 180
+SAT_OPACITY = 0.6
+
+
+def composite_channels(channel_arrays: list[np.ndarray]) -> np.ndarray:
```
braddf: Function name better as a descriptive verb to let us know that this is creating/constructing the composite layers

braddf: Or flatten even

src/quartz_api/internal/service/satellite/_ingest.py
```diff
@@ -217,6 +277,36 @@ def _run_ingest(sat_type: str) -> tuple[str, str]:
...
 
+        # Second loop to process composites
+        for name, members in COMPOSITE_CONFIG.items():
+            composite_key = f"layers/{name}/{ts_str}.tif"
+            if s3_client.object_exists(s3_bucket, composite_key):
+                continue
+
+            member_keys = [f"layers/{m}/{ts_str}.tif" for m in members]
+            if not all(s3_client.object_exists(s3_bucket, k) for k in member_keys):
+                log.info("Skipping composite %s @ %s: missing member(s)", name, ts_str)
```
braddf: Hmm interesting, I think there's an argument for still creating the composite if a later is missing, it could still be 6 layers, and technically the user could see from the independent channels that one is missing... but in an ideal world we would have a way of communicating this, which is tricky as it is at the moment.

src/quartz_api/internal/service/satellite/router.py
```diff
@@ -82,6 +82,52 @@ def get_historic_satellite_data_url(
...
+        raise HTTPException(
+            status_code=400,
+            detail=f"Invalid channel. Must be one of {sorted(VALID_CHANNELS)}",
+        )
+    if start.tzinfo is None:
+        start = start.replace(tzinfo=UTC)
+    if end.tzinfo is None:
+        end = end.replace(tzinfo=UTC)
+    if start > end:
+        raise HTTPException(status_code=400, detail="start must not be after end")
```
braddf, line 113: Is there a max date range here? I can't see anything, and we def don't want someone randomly requesting 3 months of satellite URLs 😅

src/quartz_api/internal/service/satellite/router.py
```diff
@@ -82,6 +82,52 @@ def get_historic_satellite_data_url(
     return HistoricSatelliteData(url=s3_client.get_presigned_url(bucket, key))
 
 
+@router.get("/history", response_model=list[HistoricSatelliteDataEntry])
```
braddf, line 87: Would be good to cache this for at least a short period as well I think?

braddf (changes requested):
Few comments, otherwise good!

src/quartz_api/internal/service/satellite/router.py
suvanbanerjee: The bucket only stores the past 3 days of data, but good point. I’ll add the check here as well.

src/quartz_api/internal/service/satellite/_ingest.py
suvanbanerjee: Interesting point. I think that’s definitely possible. We could add some extra metadata to the TIFF, which the FE could decode and use to display an additional message.

src/quartz_api/internal/service/satellite/router.py
suvanbanerjee: Added a 15 min cache

suvanbanerjee:
@braddf Thanks for the review! All the points have been addressed.

src/quartz_api/internal/service/satellite/cache.py
```diff
@@ -1,28 +1,116 @@
...
+
+async def warm_all_satellite_caches(
+    s3_client: S3Client,
+    channels: frozenset[str],
+    *,
+    delay: float = 5,
+) -> None:
+    """Warm the presigned-URL cache for every channel, concurrently."""
+    await asyncio.sleep(delay)
+    results = await asyncio.gather(
```
braddf, line 109: So it says in the warm function docblock that this is run in the thread to avoid being blocking, so just double checking this means that this cache function is non-blocking as it is here too? If similar to the existing v1 cache then should be but good to confirm!

braddf (approved):
One comment to check but otherwise looks great.

PR #371, opened 2026-08-17
fix: update forecaster selection logic

katyalmohit:
# Pull Request

## Description

- updated forecaster selection logic

Fixes #https://github.com/openclimatefix/client-private/issues/590

## How Has This Been Tested?

- tested with dev data-platform locally

## Checklist:

### DIFF
--- a/src/quartz_api/internal/backends/dataplatform/client.py
+++ b/src/quartz_api/internal/backends/dataplatform/client.py
@@ -142,7 +142,7 @@ class StorageClient(models.StorageInterface):
 
         if forecaster_name is None:
             # Pick whichever forecaster wrote most recently, as of now rather than the
-            # horizon cutoff, ties settled by name — so changing horizon can't change model.
+            # horizon cutoff, so that changing horizon can't change the model.
             req = messages_pb2.GetLatestForecastsRequest(
                 location_uuid=str(location_uuid),
                 energy_source=energy_type_map[energy_type],
@@ -151,9 +151,13 @@ class StorageClient(models.StorageInterface):
             resp = await self.dpc.GetLatestForecasts(req)
             if len(resp.forecasts) == 0:
                 return []
+            # Sort newest first by write time, then by init time, then by name.
+            # Two models can write at the same moment, and the data platform does not
+            # guarantee an order, so the extra keys keep the choice stable between requests.
             resp.forecasts.sort(
                 key=lambda f: (
                     -f.created_timestamp_utc.ToDatetime(tzinfo=dt.UTC).timestamp(),
+                    -f.initialization_timestamp_utc.ToDatetime(tzinfo=dt.UTC).timestamp(),
                     f.forecaster.forecaster_name,
                 ),
             )
--- a/src/quartz_api/internal/backends/dataplatform/client.py
+++ b/src/quartz_api/internal/backends/dataplatform/client.py
@@ -93,8 +93,10 @@ class StorageClient(models.StorageInterface):
     ) -> list[models.PredictedGenerationValue]:
 
         # Limit the creation time if not set
+        forecaster_pivot = created_cutoff
         if created_cutoff is None:
-            created_cutoff = dt.datetime.now(tz=dt.UTC) - dt.timedelta(
+            forecaster_pivot = dt.datetime.now(tz=dt.UTC)
+            created_cutoff = forecaster_pivot - dt.timedelta(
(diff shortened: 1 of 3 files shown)


src/quartz_api/internal/backends/dataplatform/client.py
```diff
@@ -139,19 +141,21 @@ async def get_predicted_generation(
...
-                pivot_timestamp_utc=created_cutoff,
+                pivot_timestamp_utc=forecaster_pivot,
             )
             resp = await self.dpc.GetLatestForecasts(req)
             if len(resp.forecasts) == 0:
                 return []
             resp.forecasts.sort(
-                key=lambda f: f.created_timestamp_utc.ToDatetime(tzinfo=dt.UTC),
-                reverse=True,
+                key=lambda f: (
```
peterdudfield, line 158: could you add a comment here, about what the sorting method is

peterdudfield: actually why do you need `f.forecaster.forecaster_name` in here

src/quartz_api/internal/service/regions/router.py
```diff
@@ -36,6 +36,12 @@
 
 _REGION_OBSERVER_NAME = "ruvnl"
 
+# Pinned per source, so every horizon of a chart is served by the same model.
+_REGION_FORECASTER_NAMES: dict[str, str] = {
+    "solar": "pvnet_india",
+    "wind": "windnet_india_mo_v2",
+}
```
peterdudfield, line 43: this looks good

src/quartz_api/internal/service/regions/router.py
```diff
@@ -136,8 +142,9 @@ async def get_historic_timeseries_route(
                 models.EnergyType.WIND if source == "wind" else models.EnergyType.SOLAR
             ),
             location_type=models.LocationType.REGION,
-            window_start=pd.Timestamp.utcnow().floor("H").to_pydatetime()
-            - dt.timedelta(days=2),
+            window_start=(pd.Timestamp.utcnow() - dt.timedelta(days=2))
+            .floor("D")
+            .to_pydatetime(),
```
peterdudfield, line 147: nice

src/quartz_api/internal/service/regions/router.py
```diff
@@ -205,16 +212,18 @@ async def get_forecast_timeseries_route(
     try:
         pgvs = await db.get_predicted_generation(
             location_uuid=location.uuid,
-            window_start=pd.Timestamp.utcnow().floor("H").to_pydatetime()
-            - dt.timedelta(days=2),
-            window_end=pd.Timestamp.utcnow().floor("H").to_pydatetime()
+            window_start=(pd.Timestamp.utcnow() - dt.timedelta(days=2))
+            .floor("D")
+            .to_pydatetime(),
+            window_end=pd.Timestamp.utcnow().floor("h").to_pydatetime()
```
peterdudfield, line 218: nice

src/quartz_api/internal/service/regions/router.py
```diff
@@ -278,8 +287,9 @@ async def get_forecast_csv(
 
     pgvs = await db.get_predicted_generation(
         location_uuid=location.uuid,
-        window_start=pd.Timestamp.utcnow().floor("h").to_pydatetime()
-        - dt.timedelta(days=2),
+        window_start=(pd.Timestamp.utcnow() - dt.timedelta(days=2))
+        .floor("D")
+        .to_pydatetime(),
```
peterdudfield, line 292: good

src/quartz_api/internal/backends/dataplatform/client.py
katyalmohit: > could you add a comment here, about what the sorting method is

Done

katyalmohit: > actually why do you need `f.forecaster.forecaster_name` in here

The data-platform doesn't guarantee an order in the response. So, if there are two forecasters with same `created_timestamp` it would choose the one data-platform returned first. I've now added `init_timestamp` as well to make it more deterministic. But if the `created_timestamp` and `init_timestamp`, both happened to be the same, then only it would consider the name. So, `forecaster_name` is only there to settle an exact tie.

It is arbitrary, that I agree. But, I couldn't find a better key on the response. Is there something else we could use here?

I've also added a test covering all three keys.

peterdudfield (approved):

Issue #374, opened 2026-08-21
Test cache isolation: FastAPICache.init() no-ops, so the cache leaks between tests


braddf:
`FastAPICache.init()` starts with `if cls._init: return`. The first call in a pytest session wins and every later call is a silent no-op.

Two consequences:

1. `_create_server()` re-inits the cache per test, but that does nothing after the first test — so the in-memory store is shared for the whole session and a cached route can serve a previous test's response. Hit this while adding coverage for `GET /v0/solar/GB/status`: the second test got the first one's payload. Existing tests hid it because their second case 404s, and exceptions aren't cached.

2. The ~25 `FastAPICache.init(InMemoryBackend(), prefix="test")` / `prefix="cold"` calls in `src/quartz_api/internal/service/v1/test_router.py` are all no-ops. They neither set the prefix nor provide a fresh backend. The cold-cache 503 tests are not actually starting from a cold cache — they pass on session ordering rather than on the isolation they look like they're asserting.

Fix: autouse fixture in `src/quartz_api/internal/service/conftest.py` clearing the backend between tests, then drop the no-op `init` calls in `test_router.py`.

```python
@pytest_asyncio.fixture(autouse=True)
async def clear_cache():
    if FastAPICache._init:
        await FastAPICache.clear()
    yield
```

Verified: with this in the shared conftest the full unit suite passes (223 passed) with no exceptions needed — including the v1 cold-cache tests, which get a genuinely cold cache rather than an accidental one. `FastAPICache.reset()` is also available if a test wants a full re-init.

Interim: `test_status_router.py` carries a local copy of this fixture. Remove it as part of the fix.

braddf:
Worth pinning down, since "clear the cache in every test" sounds like it should break the tests that exercise caching:

The fixture clears *before* each test and then yields. Nothing clears mid-test, so within a single test the cache is fully live — populate on the first call, serve from cache on the second.

Checked this rather than assumed it. A throwaway test against the (now cached) `/v0/solar/GB/status` route, with the fixture active, issuing two identical requests: upstream was hit once and the second response was the first one's cached body. Passes.

The existing cache-touching tests don't need cross-test state either. The v1 period tests seed their keys inside the test (`_set_forecast_meta()`, `_set_fixed_region_values()`, ...) and then read them back, and the cold-cache tests want an empty cache — which is what the fixture actually gives them, unlike the current no-op `init` calls.

CodeVishal-17:
[...]

PR #375, opened 2026-08-24
Proxy GB status for v0 from new Status API endpoints

braddf:
# Pull Request

## Description

Repoint `GET /v0/solar/GB/status` at the Quartz Status API instead of the legacy `nowcasting_datamodel` Postgres DB.

Status is now owned by the standalone Status API, which stores product status in its own colocated Postgres and serves it as the source of truth. The v0 route keeps its URL and its response shape, so no client breaks — it just sources the value over HTTP from `{STATUS_API_URL}/products/gb-solar` rather than running `SELECT * from status order by created_utc desc limit 1` and reading columns positionally.

This is an **interim step**. New integrations should query `https://status.quartz.energy/products/gb-solar` directly; the route is a temporary proxy and gets retired once clients have migrated.

**⚠️ This PR is the last piece before the legacy DB retirement, and needs an env var set before it merges.** Both are covered below.

**Scope**
- v0 only — no `/v1/{country}/{source}/status` route, v1 clients go straight to the Status API.
- GB only. Product key is hardcoded to `gb-solar`; `nl-solar` and `asset-solar` are recorded in a comment for whoever needs them.
- No config plumbing — the route is GB-only and sunsetting, so a `country_config.py` lookup would have returned a constant. Base URL is a plain `os.getenv`, read at module scope in the router, matching `time_utils.INTRADAY_LIMIT_HOURS` and `quartzdb.client.ERROR_GENERATION_CAPACITY_FACTOR`. Nothing reads the `server.conf` tree from a router today, so declaring `STATUS_API_URL` there would be documentation only — and would put the default in two places with one of them dead. Wiring the config tree into modules properly is tracked separately.

**Degrades, never 5xxs.** Any transport error, non-2xx or malformed payload returns HTTP 200 with `{"status": "unknown", "message": "Status information is currently unavailable."}`. A status endpoint that errors during an incident fails at exactly the wrong moment. This also covers the Status API's `503 status store unavailable`, which means its database is down while the API itself stays up.

Newly cached at 10s with a 3s timeout — the route was previously uncached and every hit is now an outbound call, so the cache is there to stop a burst of clients turning into a burst of outbound requests. 10s (down from 60s, per @peterdudfield) keeps that protection while making a status change visible almost immediately: this is the route people hit *during* an incident, where staleness costs more than the extra calls. It sits just inside upstream's own `Cache-Control: max-age=15`, so we re-fetch slightly more often than the Status API considers necessary — deliberate, and cheap at this volume.

Side benefit: this removes the module-level `create_engine(DB_URL)`, the last direct SQL usage in the `uk_national` service. Previously an unset `DB_URL` left `engine` undefined and raised `NameError` at request time.

## Client-facing change: new `info` status value

Status API 0.2.0 widens the status union to `ok | info | warning | error | unknown`.

`info` means a deliberate, non-degraded notice — planned maintenance, a heads-up banner. It is distinct from `unknown`, which means no signal at all. Severity order used by the Status API's own rollup is `ok < info < unknown < warning < error`.

No mapping code was needed: the v0 `Status` model types `status` as a plain `str`, so the value passes through untranslated. There is a test pinning that, because narrowing the field later would silently drop it.

**v0 consumers that switch on `ok`/`warning`/`error` will start seeing a fourth value** the first time someone sets `info` in the Status API admin. This route has never emitted it before. The field description in the v0 OpenAPI docs now enumerates the full set and tells clients to treat an unrecognised value as non-critical rather than failing.

## Merge ordering — this PR must deploy before the legacy DB is retired

On `main`, `status_router.py` is the only non-test reader of `DB_URL` in the whole service — `git grep -nE "DB_URL|create_engine" main -- 'src/**/*.py'`, the sole other hit being a test fixture in `backends/quartzdb/conftest.py`.

So the order is a hard constraint:

1. Set `STATUS_API_URL` on every deployment env
2. Merge and deploy this
3. *Then* we can retire the legacy DB → 🎉

Retire the DB first and `GET /v0/solar/GB/status` throws on every request — `engine.connect()` against a table that no longer exists. This PR is what removes that last dependency.

## Before merge: `STATUS_API_URL`

**Not currently set anywhere on this service.** It needs adding to the deployment env (Terraform / ECS task definition) per environment.

The `os.getenv` default is the *production* status URL, so a missing var does not fail loudly — dev and staging will quietly read prod's status. That's the main thing to get right before this goes out.

Locally: `STATUS_API_URL=https://status-dev.quartz.energy`.

## Commits

- `2e42ebb` — the proxy itself
- `c680c17` — response to Status API 0.2.0 (`info`); kept separate as it isn't part of the original change
- `630ecf2` — review fix: catch `TypeError` on a non-object payload

## Testing

`get_status` had zero coverage before this. Eight tests added, mocking the outbound call with `httpx.MockTransport`: upstream status and message proxied through, `info` passed through unchanged, null message coerced to `""`, upstream 500 degrades to `unknown`, connection error degrades to `unknown`, and a 200 carrying a non-object body (`[]`, `null`, a bare string) degrades to `unknown` rather than raising `TypeError`.

Verified end-to-end against the real dev Status API (`{"status": "ok", "message": "Operating within normal parameters."}`) and against a dead host (HTTP 200, `status: unknown`). Both with `DB_URL` unset, which would previously have raised `NameError`.

Note the tests carry a local autouse `FastAPICache.clear()` fixture. `FastAPICache.init()` begins `if cls._init: return`, so the in-memory store is shared across a whole pytest session and a cached route can serve a previous test's response — found while writing these. Tracked separately in #374; remove the local fixture when that is fixed.

## Checklist:


src/quartz_api/internal/service/uk_national/status_router.py
```diff
@@ -1,49 +1,64 @@
...
-
-db_url = os.getenv("DB_URL", None)
-if db_url is not None:
-    engine = create_engine(db_url)
+# Interim: status is proxied from the Quartz Status API until clients are moved
+# onto it directly, at which point this route can be retired.
+# Products are "gb-solar", "nl-solar" and "asset-solar" (sites/assets);
+# only GB has a v0 route, so the product key is fixed here rather than configured.
+STATUS_URL = os.getenv("STATUS_API_URL", "https://status.quartz.energy")
+STATUS_PRODUCT = "gb-solar"
```
suvanbanerjee, line 27: I think we might have to add it to `server.conf ` as well, and load it using conf.get_string("status-api-url") ? as this is how other env vars are currently loaded

src/quartz_api/internal/service/uk_national/endpoint_types.py
```diff
@@ -296,8 +296,18 @@ class NationalForecast(Forecast):
 class Status(EnhancedBaseModel):
     """Status Model for a single message."""
 
-    status: str = Field(..., description="Status description")
-    message: str = Field(..., description="Status Message")
+    status: str = Field(
```
suvanbanerjee, line 299: Since status has a fixed set of possible values, I think using an `enum` would give us better type safety. It might be a bit overkill though, so I'd be interested to hear your thoughts?

src/quartz_api/internal/service/uk_national/status_router.py
```diff
@@ -1,49 +1,64 @@
...
+            response.raise_for_status()
+            payload = response.json()
+        return Status(status=payload["status"], message=payload.get("message") or "")
+    except (httpx.HTTPError, ValueError, KeyError):
+        log.exception(
+            "Could not fetch status for %s from %s",
+            STATUS_PRODUCT,
+            STATUS_URL,
+        )
+        return Status(
```
suvanbanerjee, line 58: Would be good if we could send this to Sentry as well, so we know when the Status API is down. But if this is already handled elsewhere, we can omit it here.

src/quartz_api/internal/service/uk_national/status_router.py
```diff
@@ -1,49 +1,64 @@
...
-        status = Status(
-            status=row[2],
-            message=row[3],
+    try:
+        async with httpx.AsyncClient(timeout=3.0) as client:
+            response = await client.get(f"{STATUS_URL}/products/{STATUS_PRODUCT}")
+            response.raise_for_status()
+            payload = response.json()
+        return Status(status=payload["status"], message=payload.get("message") or "")
+    except (httpx.HTTPError, ValueError, KeyError):
```
suvanbanerjee: If we don't get a valid response from the Status API, for example [] or null, we could get a TypeError that isn't currently handled. Maybe we should return unknown for those cases as well.

src/quartz_api/internal/service/uk_national/endpoint_types.py
peterdudfield: the pydantic way to do this might be with a `validator`, but as @suvanbanerjee says, this could be overkill / this could be done later

src/quartz_api/internal/service/uk_national/status_router.py
peterdudfield: Thats a good idea @suvanbanerjee

src/quartz_api/internal/service/uk_national/status_router.py
```diff
@@ -1,49 +1,64 @@
...
+STATUS_URL = os.getenv("STATUS_API_URL", "https://status.quartz.energy")
+STATUS_PRODUCT = "gb-solar"
 
 
 @router.get(
     "",
     response_model=Status,
     status_code=status.HTTP_200_OK,
 )
+@cache(key_builder=key_builder, expire=60)
```
peterdudfield: could you reduce this to 10 seconds, I have when being on call, when you set the status on the analysis dashbaord, it is useful to see it pop up in the UI. If the cache time is too long, then it gives a false impression that is hasnt work

src/quartz_api/internal/service/uk_national/endpoint_types.py
braddf: I'd actually say not here; a strict enum on a response model turns an unrecognised upstream value into a ValidationError → 500, on an endpoint whose design is to never 5XX unhelpfully. 
I already did this to myself in the course of these changes, because Status API 0.2.0 added `info`, so the value set can demonstrably move underneath us — an enum stops the value being passed through to be tested by the existing logic/shown.
The field description enumerates the known set and tells clients to treat anything unrecognised as non-critical, which feels sensible for a proxy. `test_get_status_passes_through_info` pins the pass-through behaviour so nobody squishes/tightens it by accident later

src/quartz_api/internal/service/uk_national/status_router.py
braddf: Good thinking — there's a commented TODO saying the same thing, in time_utils.py:11 from Feb 2026:

```
# TODO this would be nice if this was done with the high level quartz-api config file
```

So the idea is there and I did scratch my head at this one; conf is created at the bottom of cmd/main.py, so importing it from a router is a circular import back into main, and a layering inversion on top. Not cool. 
Every module that needs an env var works around that the same way: time_utils.INTRADAY_LIMIT_HOURS, quartzdb.client.ERROR_GENERATION_CAPACITY_FACTOR, eclipse.ECLIPSE_DATE are all module-level os.getenv. This works, it just means we have mixed methods and non-complete docs of what env vars we actually need/use.

I started off the same without any HOCON, then tried a half-measure  – declare STATUS_API_URL in server.conf for discoverability, keep the os.getenv read – and then I've just removed again. Nothing reads the config tree from a router, so the block would be documentation only, with the default written in two places and only one of them connected to anything. Someone edits server.conf, nothing changes, developer swears at code, etc. etc.

The real fix is probably to move the `ConfigFactory.parse_file` call into its own module so both `main` and everything else can import `conf` with no cycle. I've added issue #376 for this, and it also clears Peter's TODO. Keeping it out of this PR because it touches `main.py` and needs its own review, whereas this one is a sunsetting v0 route.

Worth flagging on that issue though, status_router and time_utils both retire with v0, and v1 has no `os.getenv` at all – so it's more about setting/fixing the pattern properly before v1 needs its first env var in situ rather than fixing v0

src/quartz_api/internal/service/uk_national/status_router.py
braddf: Already covered elsewhere, so shouldn't need here. `main.py` calls Sentry with default integrations, which captures ERROR and above as Sentry events 👍

suvanbanerjee (approved):

PR #379, opened 2026-09-03
Finalise/simplify GB model names; make the trend adjuster a boolean

braddf:
# Pull Request

## Description

Rename the user-facing GB forecast models to describe their **input sets** rather than internal model names, so they line up across products, and move the trend adjuster out of the model name into an `adjusted` boolean. Internal DP forecaster names are unchanged — this is a naming and parameter change only.

| `model` | Inputs | Internal DP forecaster | Retired names still accepted |
|---|---|---|---|
| `blend` | blend of all models | `blend` | `blend_adjust` |
| `ecmwf_mo` | ECMWF + Met Office (day-ahead: NWPs only) | `pvnet_day_ahead` | `pvnet_day_ahead`, `pvnet_day_ahead_adjust` |
| `ecmwf_mo_sat_8h` | ECMWF + Met Office + satellite (~8h) | `pvnet_v2` | `pvnet_intraday`, `pvnet_intraday_adjust` |
| `ecmwf` | ECMWF only | `pvnet_ecmwf` | `pvnet_ecmwf`, `pvnet_ecmwf_adjust` |
| `sat_8h` | satellite only (~8h) | `pvnet_sat_only` | `pvnet_sat`, `pvnet_sat_adjust` |
| `mo` | Met Office only | `pvnet_ukv_only` | `pvnet_ukv`, `pvnet_ukv_adjust` |

National exposes all six; GSP exposes `blend`, `ecmwf_mo_sat_8h` and `ecmwf_mo`.

**`adjusted` boolean, defaulting to on** — `?adjusted=true|false` on `/regions/{id}/forecast`, `/forecast/last-updated` and `/forecasts/snapshot`. It is orthogonal to model choice and every adjustable model has an adjusted variant, so defaulting it on means callers get the best available forecast without knowing to ask. Leaves room to become a select later, or to disappear if the adjustment moves into the model pipeline.

Not every region type has adjusted variants — GB `gsp` and NL `province` have none. There the param is ignored rather than rejected: a 400 would force a caller sweeping several region types to special-case GSP. `GET /region-types` now returns `supports_adjusted` per type so that stays discoverable instead of silent.

`model_name` in responses no longer carries an `_adjust` suffix, matching v0, which stripped it before returning it too.

**Scope**
- GB naming only. NL slugs are unchanged; NL picks up the `adjusted` param and the alias mechanism for free.
- Not on `/forecasts/period` or `/generation/period` — those are cache-only and serve the pre-warmed default (adjusted); adding the param would mean warming two variants per region. 

## Backwards compatibility

The retired names in the table above are carried on `ForecastModel.aliases` / `adjust_aliases` and stay out of `/region-types`, the OpenAPI enum and 400 messages, so existing integrations keep working while new ones only ever see the current names. An alias pins its own adjuster state — the `_adjust` ones pin it on, the rest pin it off — so an old name returns byte-for-byte what it did before.

**⚠️ One behaviour change.** `blend` is both a current name and a pre-rename name, so it cannot carry an alias override: it takes the new `adjusted=true` default and now returns the adjusted blend where it previously returned the plain one. `?adjusted=false` restores the old result. Our one external tester is on `blend`, so this wants a heads-up before it ships.

## Also included

If a region type restricted intraday users but named no `intraday_default_model`, an intraday-only caller who omitted `model` got no forecaster filter at all — we sent nothing and let the data platform choose(?) or perhaps reject the request. The fallback now stays inside the intraday set, and the config invariant test requires an intraday default whenever `intraday_models` is set.

Both GB region types name one atm, so not a "now" problem but good to tighten up now rather than spotting later and having to hotfix.

## Checklist:

### DIFF
--- a/src/quartz_api/internal/service/v1/country_config.py
+++ b/src/quartz_api/internal/service/v1/country_config.py
@@ -257,8 +257,9 @@ class FM:
     NL_UNCURTAILED = ForecastModel(
         name="nl_regional_pv_ecmwf_mo_sat_uncurtailed",
         label="PV + ECMWF + Met Office + Satellite, Uncurtailed",
-        slug="ecmwf_mo_sat_uncurtailed",
+        slug="ecmwf_mo_pv_sat_uncurtailed",
         adjust_name="nl_regional_pv_ecmwf_mo_sat_uncurtailed_adjust",
+        aliases=("ecmwf_mo_sat_uncurtailed",),
         adjust_aliases=("ecmwf_mo_sat_uncurtailed_adjust",),
     )
 
--- a/src/quartz_api/internal/service/v1/country_config.py
+++ b/src/quartz_api/internal/service/v1/country_config.py
@@ -16,16 +16,48 @@ class ForecastModel:
     `name` is the internal DP forecaster_name sent to the data platform.
     `slug` is the user-facing API name (defaults to `name` when not set).
     `label` is the human-readable display name.
+    `adjust_name` is the trend-adjusted variant the `adjusted` param selects; None
+    means there is none and the param is a no-op here.
+
+    `aliases` and `adjust_aliases` are retired user-facing names, still accepted but
+    absent from `/region-types`, the OpenAPI enum and error messages. `aliases` pins
+    the adjuster off and `adjust_aliases` pins it on, so an old name keeps the exact
+    forecast it returned before the adjuster became a parameter.
     """
 
     name: str
     label: str
     slug: str | None = None
+    adjust_name: str | None = None
+    aliases: tuple[str, ...] = ()
+    adjust_aliases: tuple[str, ...] = ()
 
     @property
     def api_name(self) -> str:
         """User-facing model name used in API params and enum values."""
         return self.slug if self.slug is not None else self.name
+
+    def accepts(self, api_name: str) -> bool:
+        """True if `api_name` names this model, current or legacy."""
+        return (
+            api_name == self.api_name
+            or api_name in self.aliases
+            or api_name in self.adjust_aliases
+        )
+
+    def alias_adjust_override(self, api_name: str) -> bool | None:
+        """Trend-adjuster state implied by a legacy name, or None if it implies nothing."""
+        if api_name in self.adjust_aliases:
+            return True
+        if api_name in self.aliases:
+            return False
+        return None
+
+    def internal_name(self, *, adjusted: bool) -> str:
+        """Internal DP forecaster_name for this model at the given adjuster setting."""
+        if adjusted and self.adjust_name is not None:
+            return self.adjust_name
+        return self.name
 
 
 @dataclass(frozen=True)
@@ -38,8 +70,10 @@ class RegionTypeConfig:
     location_type: LocationType
     source_types: tuple[str, ...] = ()
     forecast_models: tuple[ForecastModel, ...] = ()
-    # default_model stores the internal DP forecaster_name, not the user-facing slug.
+    # Internal DP forecaster_name, unadjusted — the `adjusted` param reaches the variant.
     default_model: str | None = None
+    # False (e.g. GB gsp) means no adjusted variants exist, so `adjusted` is a no-op.
+    supports_adjusted: bool = False
     metadata_fields: tuple[str, ...] = ()
     # intraday_models lists models accessible to intraday-only users (subset of forecast_models).
     intraday_models: tuple[ForecastModel, ...] = ()
@@ -56,18 +90,29 @@ class RegionTypeConfig:
         return None
 
     def get_model_by_api_name(self, api_name: str) -> ForecastModel | None:
-        """Look up a ForecastModel by its user-facing API name (slug or internal name)."""
+        """Look up a ForecastModel by its user-facing API name, current or legacy."""
         for fm in self.forecast_models:
-            if fm.api_name == api_name:
+            if fm.accepts(api_name):
                 return fm
         return None
 
     def get_model_by_internal_name(self, name: str) -> ForecastModel | None:
-        """Look up a ForecastModel by its internal DP name."""
+        """Look up a ForecastModel by its internal DP name, adjusted or not."""
         for fm in self.forecast_models:
-            if fm.name == name:
+            if name in (fm.name, fm.adjust_name):
                 return fm
         return None
+
+    def default_forecaster_name(self, *, adjusted: bool = True) -> str | None:
+        """Internal DP forecaster_name for the default model at the given adjuster setting."""
+        if self.default_model is None:
+            return None
+        fm = self.get_model_by_internal_name(self.default_model)
+        if fm is None:
+            return self.default_model
+        return fm.internal_name(
+            adjusted=adjusted and self.supports_adjusted,
+        )
 
     def default_model_api_name(self) -> str | None:
         """User-facing API name for the default forecast model."""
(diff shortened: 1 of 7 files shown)


src/quartz_api/internal/service/v1/country_config.py
```diff
@@ -16,17 +16,49 @@ class ForecastModel:
...
 
     @property
     def api_name(self) -> str:
         """User-facing model name used in API params and enum values."""
         return self.slug if self.slug is not None else self.name
 
+    def accepts(self, api_name: str) -> bool:
+        """True if `api_name` names this model, current or legacy."""
+        return (
+            api_name == self.api_name
```
peterdudfield: I think you can do `api_name in [self.api_name, *self.aliases, *self.adjust_aliases]`. to make it neater

peterdudfield:
What country is this for?

```
⚠️ One behaviour change. blend is both a current name and a pre-rename name, so it cannot carry an alias override: it takes the new adjusted=true default and now returns the adjusted blend where it previously returned the plain one. ?adjusted=false restores the old result. Our one external tester is on blend, so this wants a heads-up before it ships.
```

peterdudfield:
Would anything in the api checks [here](https://github.com/openclimatefix/airflow-dags/blob/main/src/airflow_dags/dags/uk/check-nl-api.py) need to change?

src/quartz_api/internal/service/v1/country_config.py
```diff
@@ -16,17 +16,49 @@ class ForecastModel:
...
+    absent from `/region-types`, the OpenAPI enum and error messages. `aliases` pins
+    the adjuster off and `adjust_aliases` pins it on, so an old name keeps the exact
+    forecast it returned before the adjuster became a parameter.
     """
 
     name: str
     label: str
     slug: str | None = None
+    adjust_name: str | None = None
+    aliases: tuple[str, ...] = ()
```
peterdudfield, line 32: how come you didnt make these lists of strings?

peterdudfield (approved):
Ive approved, 

Ive add a few questions, but nothing major

src/quartz_api/internal/service/v1/routes/forecasts.py
```diff
@@ -196,7 +212,7 @@ async def get_forecast_last_updated_timestamp(
         )
     location_type = locs[0].location_type or models.LocationType.NATION
     rt = country.location_type_to_region_type(location_type)
-    model = resolve_forecast_model(model, rt, is_intraday_only)
+    model = resolve_forecast_model(model, rt, is_intraday_only, adjusted)
```
suvanbanerjee, line 216: I think before this add `validate_model(model_name, rt, rt.type)`

src/quartz_api/internal/service/v1/country_config.py
```diff
@@ -16,17 +16,49 @@ class ForecastModel:
...
     def api_name(self) -> str:
         """User-facing model name used in API params and enum values."""
         return self.slug if self.slug is not None else self.name
 
+    def accepts(self, api_name: str) -> bool:
+        """True if `api_name` names this model, current or legacy."""
+        return (
+            api_name == self.api_name
+            or api_name in self.aliases
+            or api_name in self.adjust_aliases
```
suvanbanerjee: I think here if someone tries to get GSP with blend_adjust this will allow it, check this behavior against the old object check once

suvanbanerjee (commented):
Couple of comments from my side but overall looks solid!

src/quartz_api/internal/service/v1/country_config.py
braddf: Amended! Now checks for these "non-existent" regional models if it's an adjust_alias and returns false if they shouldn't be allowed. Good catch 👍

src/quartz_api/internal/service/v1/routes/forecasts.py
braddf: Agreed, added. It was the only forecast route not validating, potentially hitting the DP instead of returning 400, must have missed that one!

suvanbanerjee (approved):

braddf:
> What country is this for?
> 
> ```
> ⚠️ One behaviour change. blend is both a current name and a pre-rename name, so it cannot carry an alias override: it takes the new adjusted=true default and now returns the adjusted blend where it previously returned the plain one. ?adjusted=false restores the old result. Our one external tester is on blend, so this wants a heads-up before it ships.
> ```

Both GB and NL national – thanks, the description said just GB but this does apply to NL going forwards too, even if it's backward compatible for both countries for other model names. **`?model=blend` resolved to `nl_blend` before and `nl_blend_adjust` now, matching new GB behaviour.** 
NL `province` and GB `gsp` stay the same, as neither has adjusted variants nor were these allowed before with the `_adjust` suffix. One other thing I've just realised re NL naming, hopefully the last thing to finalise: `ecmwf_mo_sat_uncurtailed` isn't being renamed, so that one does change behaviour alongside the blend, but I wonder if we should actually rename this one anyway to `blend_uncurtailed`, unless that's technically not accurate..?

src/quartz_api/internal/service/v1/country_config.py
braddf: I'll admit my view of Python dataclasses rules and regs is still mixed, but I spent a bit of time thinking about this config, and there are a couple of things here trying to keep these top Classes tight as poss:

A plain `list` default won't actualy import – dataclasses reject it: `ValueError: mutable default <class 'list'> for field aliases is not allowed: use default_factory`. So every alias field would need wrapping in `field(default_factory=list)`, which feels like a bit of shoehorning in comparison to a tuple, which does work.

Secondly, `frozen=True` only stops you *replacing* a field, not changing what's inside it. `fm.aliases = (...)` is blocked, but `fm.aliases.append(...)` would work fine on a list. `COUNTRIES` is built once when the module loads and shared by every request, so a stray `append` anywhere would change model routing for everything that worker handles afterwards... and only that worker, which would be horrible to debug. A tuple has no `append`, so it can't happen by design, so I'd say this is the pattern we want to adopt

src/quartz_api/internal/service/v1/country_config.py
braddf: Yep updated, used a tuple instead of a list so we're not building a throwaway list on every call, but same idea.

Then I actually dropped this `accepts()` altogether. It only had one caller, and then from tweaks re Suvan's comment, that caller always passed in the region type's own `supports_adjusted` flag just to get it handed back – so the check now lives in `get_model_by_api_name`, right next to `get_model_by_internal_name`, which already reads those fields anyway.

braddf:
> Would anything in the api checks [here](https://github.com/openclimatefix/airflow-dags/blob/main/src/airflow_dags/dags/uk/check-nl-api.py) need to change?

Had a look – nothing needs to change. `check-nl-api.py` never passes a model: it's left as `None` at both `check_forecast` call sites, and `check_forecast_at_timestamp` uses `province`, whose default is unchanged as well 👍 

Separate thing I noticed while I was in there, in the DAGs repo rather than this one – [`check-nl-api.py` line 70](https://github.com/openclimatefix/airflow-dags/blob/7fa96caceec169d48d2621b741fe8d871743b38e/src/airflow_dags/dags/uk/check-nl-api.py#L70) adds `model_name=...` to the per-region forecast URL, but that route's parameter is called `model` (`model_name` only exists on `/forecasts/snapshot`, and that's annoying so we should line it up to match the rest of the routes too – I have a [separate issue](https://github.com/openclimatefix/quartz-api/issues/380) to go through and make sure we fix all these tiny inconsistencies). Nothing passes it at the mo, so it's harmless so far, but anyone who adds a check with that will get it silently ignored and probably be confused!

braddf:
cc. @peterdudfield in case you miss these comments ☝️

src/quartz_api/internal/service/v1/country_config.py
peterdudfield: Thanks, that sounds sensible

peterdudfield:
> > Would anything in the api checks [here](https://github.com/openclimatefix/airflow-dags/blob/main/src/airflow_dags/dags/uk/check-nl-api.py) need to change?
> 
> Had a look – nothing needs to change. `check-nl-api.py` never passes a model: it's left as `None` at both `check_forecast` call sites, and `check_forecast_at_timestamp` uses `province`, whose default is unchanged as well 👍
> 
> Separate thing I noticed while I was in there, in the DAGs repo rather than this one – [`check-nl-api.py` line 70](https://github.com/openclimatefix/airflow-dags/blob/7fa96caceec169d48d2621b741fe8d871743b38e/src/airflow_dags/dags/uk/check-nl-api.py#L70) adds `model_name=...` to the per-region forecast URL, but that route's parameter is called `model` (`model_name` only exists on `/forecasts/snapshot`, and that's annoying so we should line it up to match the rest of the routes too – I have a separate issue to go through and make sure we fix all these tiny inconsistencies). Nothing passes it at the mo, so it's harmless so far, but anyone who adds a check with that will get it silently ignored and probably be confused!

Sounds good, do you mind adding the link of that issue here, just so it links back to this comment?

braddf:
@peterdudfield added link to my reply and also added issue for check here: https://github.com/openclimatefix/airflow-dags/issues/806

src/quartz_api/internal/service/v1/country_config.py
```diff
@@ -151,98 +195,92 @@ class FM:
...
         name="nl_regional_pv_ecmwf_mo_sat_uncurtailed",
         label="PV + ECMWF + Met Office + Satellite, Uncurtailed",
         slug="ecmwf_mo_sat_uncurtailed",
-    )
-    NL_UNCURTAILED_ADJUST = ForecastModel(
-        name="nl_regional_pv_ecmwf_mo_sat_uncurtailed_adjust",
-        label="PV + ECMWF + Met Office + Satellite, Uncurtailed (Trend Adjusted)",
-        slug="ecmwf_mo_sat_uncurtailed_adjust",
+        adjust_name="nl_regional_pv_ecmwf_mo_sat_uncurtailed_adjust",
+        adjust_aliases=("ecmwf_mo_sat_uncurtailed_adjust",),
```
peterdudfield, line 263: change slug to ="pv_ecmwf_mo_sat_uncurtailed_adjust"

aliases = "ecmwf_mo_sat_uncurtailed",

PR #381, opened 2026-09-07
fix: align /sites/{uuid}/forecast window to UTC day boundaries

PavanRaghavendraKulkarni:
# Pull Request

## Description

The endpoint was rolling its ±2-day window to the current hour (pd.Timestamp.now(tz).floor("H")), so the window start/end drifted throughout the day depending on when the request was made. The legacy India API this replaces always aligns the window to UTC midnight, so the two could disagree by up to ~24 hours on which forecast points are returned. This aligns the new endpoint to the same UTC-midnight ±2-day window, verified against the old API for all 8 Adani sites.

Fixes #

## How Has This Been Tested?

Ran the patched `get_forecast` window logic locally against the Data Platform (via gRPC) for all 8 Adani sites (`ad_AP250`, `ad_khavda_pv`, `ad_KA200`, `ad_AEML_250MW`, `ad_Seci_5`, `ad_khavda_wind`, `ad_Wind_325`, `ad_AEML-1`) and compared the output to a live call to the old (pre-migration) India API for the same sites. Before the fix, window start/end and point counts diverged by up to several hours depending on time of day; after the fix, point counts match almost exactly and values agree within single-digit-to-low-double-digit kW.

- [x] Yes

## Checklist:


PavanRaghavendraKulkarni:
Screenshot 2026-09-07 at 10 52 52 AM

peterdudfield (approved):

PR #391, opened 2026-09-14
GB → v1: Final release tweaks

### LINKED ISSUE #380: Clean up any small inconsistencies remaining (backward-compatible)
- [x] `model` vs. `model_name` as a query param across all routes
- [x] ~GSP-level forecast req for national model doesn't 403~ User error! But have made this and the two below clearer defaults
- [x] ~model select on forecast not working~
- [x] ~model select on last_updated route~
- [x] `/regions` returns a deterministic order (was unordered between cache expiries)
- [x] ensure all run details (e.g. init and last_updated) are available, correct and sensible
  - [x] `last_updated_utc` / `latest_init_utc` from the latest run in the response, not `values[0]` — on `/forecasts/snapshot` that was one arbitrary region of ~330
  - [x] Top-level `capacity_kW` from the last target time, not the first value
  - [x] `detail=values|runs|full` for per-value run metadata, and the forecaster's own metadata — closes v0's `input_data_last_updated` gap
  - [x] Descriptions on the hoisted timestamps and capacity, so which run/capacity you're looking at isn't guesswork
- [x] gRPC failures answered as HTTP errors — an unknown pinned model  version was returning a raw `AioRpcError` repr
- [x] Scalar docs: disabled parameter rows legible, and parameters ticked on input so typed values are actually sent
- [x] order on docs request window left hand side, e.g. query params near top
- [x] p-levels in order
- [x] capacity national seems low?

Check this is correct behaviour:
- [ ] No `plevels_kW` on GSPs for `blend`, but for `ecmwf_mo` and `ecmwf_mo_sat_8h ` there are p10 and p90

braddf:
# Pull Request

## Description

A number of tidy-ups to the v1 API ahead of opening it to users. Some correct response fields that were reproducible only by luck, one adds the opt-in detail that v0 clients lose without it, several are docs and error-handling fixes, and the last few close gaps found in review.

**No existing response changes shape.** No field is added, removed or renamed on a default response, and everything new is either backward-compatible or opt-in behind a parameter.

Two things do change for existing callers, so they are worth reading before approving.

Some fields change *value*. The hoisted run metadata moves from an arbitrary pick to a defined one (below). Power and capacity in kW are now rounded rather than truncated, which moves most values by under a kW (below). NL's `observer_name` now reads `ned_nl` rather than the platform's internal `nednl`.

Some requests that used to return 200 now return an error: a site or substation as a region (404), a region UUID from another country (404), `/forecasts/period` on an intraday-only subscription (403), an `end_utc` over a year ahead (422), a negative `horizon_minutes` (422), an unmatched `region_names` (400), and an observer not configured for the country (400). A request with no bearer token changes from 400 to 401. Each has its own section below.

Fixes #380 

## Sites and substations are no longer reachable as regions

The most serious thing in this PR. `GET /regions?parent=national` returned 990 entries: 337 GSPs, plus 589 primary substations and 64 individual sites, including test sites. Fetching one of those names on `/regions/{region}/forecast` returned a site forecast and named the internal `pv_site_production` forecaster.

The cause is that the platform's enclosing filter is transitive, so a lookup under a nation returns everything beneath it, and the untyped lookups did not restrict what came back. A country's configured region types are the whole of its API surface, so anything else is now dropped: 990 becomes 337, verified against a live platform.

The same untyped lookup backed region-name resolution and the UUID check above, so a site was reachable by name and by UUID as well as in the listing. All three go through one `is_api_region` check now, and the three routes that refetch a region by UUID share a `fetch_api_region` that repeats it at the point of fetch. Keeping the check in both places means resolving an id some other way cannot reopen it.

## `model_name` and `model_version`

`model` is renamed to `model_name` on all three model-selecting routes, with `model_version` alongside it to pin a specific version. The old `model` name is still accepted but kept out of the OpenAPI schema, so existing integrations keep working while new ones only ever see the current name. Supplying both with different values is a 400.

## `observer_name`

`observer` is renamed to `observer_name` on the generation routes, in the same way as `model_name` above, matching the `observer_name` field already returned in the response. The old name still works and stays out of the schema; both with different values is a 400.

Two NL-only bugs came out of it, both live. The default observer was hardcoded to `pvlive_in_day` for every country, so an NL request naming no observer asked the platform for PV Live over a Dutch province; the default now comes from `country_config`. And the snapshot route echoed the platform's internal `nednl` where the per-region route echoed the API's `ned_nl`; both now return the API name.

## Scalar docs: disabled parameter rows

Scalar leaves an optional parameter unticked when you type a value into it, so the param is excluded from the request and the caller gets the default back with no indication why. Text params are now ticked on input, disabled rows are greyed, and a disabled row that holds a value is labelled. Pins the Scalar bundle the selectors were written against to avoid these CSS tweaks going awry.

Two further fixes to the same panel. Query Parameters now sits directly under Variables instead of below Cookies and Headers, since that is what callers actually set on this API. Scalar has no configuration(?) for section order, so this sets flex `order` from the section's header text (the element ids are sequential and shift by route, so neither id nor `:nth-child` identifies a section reliably). And a ticked row's checkbox now uses the accent colour, so which rows are actually being sent are more obvious; the greying above was also extended to dropdown rows, which were still showing their value at full strength while unticked and not being sent. Both degrade to Scalar's own appearance if it renames these elements.

Third fix in the same area: the OAuth **Redirect URL** was blank and stayed blank. It is now sent on every docs page load, derived from the request so it follows the environment (`https://…/v1/docs` behind the proxy, `http://localhost:8000/v1/docs` locally) rather than being hardcoded, and it needs no new Auth0 callback registration because it is the docs page itself. If Scalar changes its storage shape nothing matches and it is a no-op and doesn't break anything else.

Fourth in the same area: the docs page could not refresh its token. Auth0 only issues a refresh token when `offline_access` is asked for, and it was requested on the v0 Swagger page but not the v1 Scalar one, so the refresh control had nothing to use and said so. All four scopes are now requested and pre-ticked, because a scope that is not ticked at login is not in the token and cannot be added to it afterwards. The scope list is defined once and reaches the security scheme, the requested scopes and the ticked scopes from there. The per-request redirect URL used to be assigned over that same config block, which would have wiped the client id and scopes now sitting in it, so it merges instead.

## gRPC failures answered as HTTP errors

An unknown pinned model version reached the caller as a raw `AioRpcError` repr, `debug_error_string` and all. gRPC statuses now map onto HTTP ones on both apps, and an unknown model/version pair returns a 404 naming what was not found.

## Deterministic region ordering

The data platform gives no ordering guarantee on `ListLocations`, so `GET /{country}/{source}/regions` could return the same regions in a different order each time the 60s cache expired. Regions now sort by region type level, then by name. National types are level 0, so they sort to the top without a special case.

## Run metadata derived from all values, not `values[0]`

A forecast response raises `last_updated_utc`, `latest_init_utc` and `capacity_kW` onto the wrapper and took all three from one arbitrary element of the result set, and on `/forecasts/snapshot` it was `snapshot[0]`, so the first *region*, not the first timestep, so on a GB GSP snapshot those fields came from one of ~330 regions chosen arbitrarily, and could differ between two identical requests.

| Field | Was | Now |
|---|---|---|
| `last_updated_utc`, `latest_init_utc` on `/regions/{region}/forecast` | first value's run | latest run across all values |
| the same, on `/forecasts/snapshot` | one arbitrary region's run | latest run across all regions |
| `capacity_kW` on forecast and generation time series | first value's capacity | capacity at the last target time |

This is the treatment `/forecasts/period`'s pre-warmer already applied so have used that across the board. `last_updated_utc` means "the most recent forecast update contributing to this response" and `latest_init_utc` should actually be accurate now.

`model_name` and `model_version` are deliberately left taking the first value's: a maximum over them is meaningless, and a single request pins one forecaster by name anyway. The more accurate/specific answer for those is per-value attribution, also in this PR below.

## `detail` – per-value run metadata

`?detail=values|runs|full` on the per-region forecast and generation routes. Additive and opt-in; omit it and the response is byte-identical to today.

| `detail` | Adds to each value |
|---|---|
| `values` (default) | nothing – `time_utc`, `power_kW`, `plevels_kW` as before |
| `runs` | `last_updated_utc`, `latest_init_utc`, `model_name`, `model_version`, `capacity_kW` |
| `full` | the forecaster's own metadata dict as well |

This is what a caller doing accuracy analysis needs and could not get: group errors by model run, detect a model version change inside a window, verify the horizon actually returned, or see a mid-window capacity change that a single hoisted number cannot express. An enum rather than a boolean so a fourth level can be added without a second parameter. It is part of the cache key, with a regression test.

**`detail=full` closes the v0 `input_data_last_updated` gap.** The forecaster metadata carries input-freshness timestamps, confirmed against live GSP rows, but the keys vary by forecaster: `blend` records `nwp_last_updated` and `satellite_last_updated`, `sat_only` records four differently-named ones. So the dict passes through as the forecaster wrote it, under the name `metadata` rather than `input_data_last_updated`, which would imply a shape we are not promising. The OpenAPI description says the keys are not a stable contract. `app_version` is parsed from the string the pipeline emits into a real object so clients don't parse twice.

Normalising to v0's `{gsp, nwp, pv, satellite}` needs a per-forecaster key map that goes stale whenever the pipeline adds a key, and would discard two of the timestamps `sat_only` reports. Once all models run through one pipeline the keys become predictable and this can become a typed structure; shipping it as an explicitly unstable passthrough behind an opt-in level is what makes that swap non-breaking.

On generation routes there is no model run, so `runs` adds only per-value capacity and `full` adds nothing further. It is accepted there so one request shape works across both routes. `detail` is deliberately absent from `/forecasts/period` and `/generation/period`, which serve pre-warmed data as a bulk "latest block" where per-value provenance is not what the endpoint is for.

## Probability levels in percentile order

`plevels_kW` came back in an arbitrary key order, from two independent causes. The data platform returns them with no ordering guarantee, and `/forecasts/period` built its key set from a set, so its order was arbitrary regardless of what the backend returned. Both are now sorted in the API layer, at all three response sites (per-region forecast, snapshot, matrix), so the ordering holds for every backend and also for cache entries that were warmed before this change. Sorting is by just the numeric part, so `p2` precedes `p10` precedes `p90` rather than the string order `p10`, `p2`, `p90`. A key that is not `p<digits>` sorts last instead of erroring, so an unexpected level from the platform is still returned.

## Installed capacity, deliberately out of the way

`capacity_kW` is the platform's *effective* capacity, a few percent below the `installedCapacityMw` v0 reported and PV Live publishes – v0 preferred the non-degraded figure from location metadata, and v1 gives us an opportunity to reassess this. Effective is what the forecast is normalised against, and the non-degraded number does not reflect reality, with the allowance that users might already be using it from v0 / PV Live, so it still could be of use (and hopefully just temporarily in some cases).

Hence the non-degraded figure is still reachable, because clients migrating from v0 have only ever had that one, but it is placed so it has to be sought out: `installed_capacity_kW` inside `RegionDetail.metadata` on the region endpoints, and inside a `metadata` object on the forecast and generation wrappers that appears only at `detail=full`. It is omitted where the DP has no value, rather than echoing the effective capacity or being erroneously 0. The field description says it is not what the forecast uses and that `capacity_kW` is preferred.

## Field descriptions

The hoisted timestamps and capacities now carry OpenAPI descriptions saying what is actually selected, so a reader does not have to guess which run or which capacity they are looking at.

## Watt-to-kW conversion

The watts-to-kW conversion was written out at thirteen call sites and had drifted: most truncated with `int()`, one truncated before dividing, one rounded to 4dp, so the same region's power differed between the timeseries and snapshot routes. All thirteen now go through one `_kw()` helper rounding to 3dp.

Truncating was also losing data. The fractions arrive from the platform as float32, so an exact `0.9` is really `0.89999997` and `int()` turned an exact 900 kW into 899, always downwards. **This changes returned values**, by under a kW. 3dp is 1 W, which is well inside the float32 noise the platform's own numbers carry.

## Access checks

`/forecasts/period` called the country access check and discarded the result, so an intraday-only subscription was served the full blend matrix. It serves one pre-warmed model per region type, so there is nothing to downgrade to: it now returns **403** pointing at the per-region route. GB is the only country with an intraday tier.

A region path segment that parsed as a UUID was returned without checking it belongs to the country in the path, so a UUID from one country resolved under another's. Since the country permission is checked against the path, that was the step between a single-country subscription and another country's data. It is now confirmed to be the nation or enclosed by it, verified against a live platform.

The admin scope on both refresh routes is the `ADMIN_PERMISSION` constant rather than the string written out twice.

## `end_utc` validation

`start_utc` was checked against the 1-year limit and `end_utc` was not checked at all, so a mistyped year came back as an empty series that looked like missing data. It is now a **422** beyond a year ahead.

This also fixes a 500: a naive `end_utc`, sent without a timezone, was compared against an aware `start_utc` and raised a `TypeError`.

## Smaller fixes from review

- `horizon_minutes=-5` reached the platform and returned a **500**. Now `ge=0`.
- A `time_utc` off the half hour returned an empty snapshot, since values only exist on the half hour. It is now floored, and the `time_utc` in the response says which timestamp was used.
- An unmatched `region_names` on the period routes returned **200** with nothing in it. It is now a **400** naming what was not found.
- Pinning an unknown `model_version` returned a 404 naming the internal forecaster (`blend_adjust`) rather than the name the caller passed (`blend`).
- Model validation errors wrote the region type as `GSP`, the internal enum name, where the API says `gsp`.
- A request with no bearer token got a **400** `invalid_request` with an empty description; it is now a **401** with a `WWW-Authenticate` challenge. Keyed on the shape of the `Authorization` header rather than on the library's status or error code, because a missing argument — a programming error on our side — arrives as the same 400 with the same `invalid_request` code. A token that is present but bad still goes to the verifier.

## Documentation corrections

The API description said regions carry a `region_name` (they carry `name`) and that region IDs come from `/regions` (it returns no ID field). `/generation-sources` said to pass `name` as `observer`. The forecast route promised 48 hours ahead where a response reaches only as far as the latest model run. `detail`'s description was one blob covering both routes, mentioning model runs on the generation route where there are none, and neither mentioned the top-level `metadata` that `full` adds. All corrected, and em dashes are out of the user-facing text.

## Spec, limits and CORS

A second review pass, all of it things the spec claimed or omitted rather than wrong behaviour, bar the last two.

The OpenAPI schema declared only `200` and `422`, so every other status a caller will actually meet was undocumented and a generated client had no type for it. Routes now declare `400`, `401`, `403`, `404`, `429` and `503` as appropriate, against one shared `ErrorDetail` model. The v1 schema also emitted `"tags": null`, and the `region` path parameter had no example, so the docs rendered curl snippets with a literal `{region}` in the URL. Rate limits (20/second, 3,600/hour, per user per route) and the 3-month window cap were both enforced and undocumented; both are now in the description.

**429 responses returned `{"error": ...}`** where every other error in the API returns `{"detail": ...}`, and carried no `Retry-After`. slowapi's stock handler is replaced with one that matches the rest of the API and says when to retry.

**CORS allowed credentials from any origin.** `allow_origins` defaults to `*` with `allow_credentials=True`, and Starlette responds to that by echoing the requesting origin back with `Access-Control-Allow-Credentials: true`. The flag is now set only when origins are named explicitly. Nothing in this API needs it either way, since it authenticates with a bearer token that a browser never attaches on its own.

**The v0 description was assembled and discarded.** `FastAPI()` is constructed before the auth branch runs and never received `description`, so both the base text and the whole authentication section were appended to a local that went out of scope. `make_api_auth_description()` has never reached the docs page.

## Testing

New tests use backend stubs returning deliberately awkward data — values spanning several runs, regions returned in reverse order, a nation enclosing sites and substations as the real platform does — and each was confirmed to fail before its fix.

They cover the hoisted run metadata being order-independent, `detail` at each level including its effect on the cache key, the per-country observer default and alias, the intraday 403 on the period route (and its absence on generation, which has no model tiering for it to leak into), a region UUID from outside the country, an `end_utc` years ahead, and sites being unreachable as regions by name, by UUID, and in the listing.

Also adds an `isolated_cache` fixture. `FastAPICache.init` returns early once any test has initialised it, and `InMemoryBackend._store` is a *class* attribute shared by every instance, so both have to be reset; without it two tests hitting the same URL had the second served the first's cached response.

## Checklist:


src/quartz_api/internal/service/v1/routes/forecasts.py
suvanbanerjee: Import this from `src/quartz_api/internal/service/v1/auth_scopes.py`

src/quartz_api/internal/service/v1/routes/generation.py
suvanbanerjee: Read it from `src/quartz_api/internal/service/v1/auth_scopes.py`

src/quartz_api/internal/service/v1/helpers.py
suvanbanerjee: I think we should add a `company_id` check here, any user if then have the site UUID can access its forecast even of they don't belong to the company 

PS - its not possible to guess a UUID by brute forcing but i think its better to make it strict

src/quartz_api/internal/service/v1/routes/forecasts.py
suvanbanerjee: I think this will not block `read:uk-intraday` to get blend it will only return false

REF -  helpers.pyL242
```
if cfg.intraday_permission and cfg.intraday_permission in perms:
        return False
```

Maybe add a `if check_country_access(auth, country)` check below this

src/quartz_api/internal/service/v1/routes/generation.py
```diff
@@ -54,6 +57,7 @@
     "/{country}/{source}/regions/{region}/generation",
```
suvanbanerjee, line 62: add a `valid_observers` check in this route

src/quartz_api/internal/service/v1/routes/generation.py
suvanbanerjee, line 154: add a `valid_observers` check in this route

src/quartz_api/internal/service/v1/helpers.py
suvanbanerjee, line 453: it would be good if we check `end` is a valid ts, `start` has a `_check_window_start` function, similar we can add a check for `end_utc`

Image

src/quartz_api/internal/backends/dataplatform/client.py
suvanbanerjee, line 1: In this file `int(capacity_watts * fraction / 1000)` or a variation is used multiple times with different precision, maybe add a common internal helper function  like `_kw` and use it everywhere

suvanbanerjee:
image
In  the docs 

Discovery ... Each region carries a **region_name** identifier used across all other endpoints.
it should be just `name` not `region_name`

Also **Region IDs** are unique strings — obtain them from the /regions endpoints.
I don't think we return the region_id

---

image

the refresh token button is not saying no refresh token available

---

image

By default the redirect url is empty

suvanbanerjee (approved):
LGTM!

src/quartz_api/internal/service/v1/helpers.py
braddf: In general this will be a bit of co-opting checks on both sides to ensure there isn't any bleed between regions and sites, so we'll handle that in the v1 Sites work 👍

braddf:
Just added a final batch of small fixes from a final Claude sweep. Should be good to go now I think!

PR #394, opened 2026-09-16
v1: docs polish + NL quarter-hour snapshots

braddf:
# Pull Request

## Description

A small v1 tidy-up, plus one regression fix for NL 🇳🇱

NL snapshots were accidentally missing alternately... Both snapshot routes rounded every time down to the half hour, but NL data from the DP is quarter-hourly, so `:15` and `:45` couldn't be requested. Each country in `country_config.py` now declares its `time_step_minutes` (GB 30, NL 15), and snapshots round to that instead. It's required, so any new country has to set it, and we've had to add a similar thing in the UI already so we might want to serve this as part of the country metadata, TBC.

**Docs polish**
- The period routes are now "for Current Period", which makes it clearer they're the live block and not a window you can move into the past.
- Added an OCF logo in the Scalar sidebar.
- Tweaked the button and link colours.
- Dropped the "30-minute resolution" wording, since that was only true for GB.

## Testing

- New tests for NL snapshots at `:45` and for the rounding in each country.
- Checked the time steps against live DP data: NL is 15 min for both forecasts and `nednl`, and GB is 30.
- Checked the NL and GB snapshots and the docs page on a local server.

## Checklist:


src/quartz_api/cmd/main.py
```diff
@@ -434,23 +434,85 @@ async def v1_scalar_docs(request: Request) -> HTMLResponse:
...
+                      /* Scalar has no logo option (`x-logo` is Redoc-only), so the
+                         logo is drawn above the sidebar search. The sidebar is a
+                         flex column, so this sits as its first item. */
+                      aside.t-doc__sidebar::before {
+                        content: "";
+                        display: block;
+                        flex-shrink: 0;
+                        height: 32px;
+                        margin: 20px 20px 8px;
+                        background: url("/static/logo-ocf.svg") left center / contain
```
suvanbanerjee, line 509: `static/logo-ocf.svg` it should be `logo.png`

src/quartz_api/internal/service/v1/country_config.py
```diff
@@ -199,6 +207,15 @@ def resolve_observer(self, api_name: str) -> str:
                 return gs.name
         return api_name
 
+    def floor_to_time_step(self, ts: dt.datetime) -> dt.datetime:
+        """Floor a timestamp to this country's time step, as UTC.
+
+        A naive timestamp is taken to be UTC already.
```
suvanbanerjee: I think it will be better if we convent to UTC as well if input is not UTC

src/quartz_api/internal/service/v1/routes/forecasts.py
suvanbanerjee: I think use the new function here too

src/quartz_api/cmd/main.py
```diff
@@ -434,23 +434,85 @@ async def v1_scalar_docs(request: Request) -> HTMLResponse:
...
+                        --scalar-color-accent: #ff4901;
+                        --scalar-color-green: #57afa8;
+                        --scalar-color-red: #dd2f2c;
+                        --scalar-color-yellow: #ffd072;
+                        --scalar-color-blue: #64b0c9;
+                        --scalar-color-orange: #f99f56;
+                        --scalar-color-purple: #4575c1;
+                        --scalar-button-1: #fffbf4;
+                        --scalar-button-1-hover: #fffbf4e6;
+                        --scalar-button-1-color: black;
```
suvanbanerjee, line 461: Here instead of defining it twice you can maybe do

```
:root .dark-mode,
:root .light-mode {
  --scalar-color-accent: #ff4901;
  ...
```

suvanbanerjee (changes requested):

src/quartz_api/cmd/main.py
braddf: Yeah these are similar but not identical atm, and when we implement light mode they may need tweaking more so I quite like the separation, but I've merged the btn styling as they're identical for now

src/quartz_api/internal/service/v1/country_config.py
braddf: Yep fair, have added in along with a test for non-standard offset

src/quartz_api/cmd/main.py
braddf: Ha, no I just forgot to stage the new SVG logo 😅 have committed properly now

suvanbanerjee (approved):

PR #400, opened 2026-09-18
v1 docs: rebrand to OCF Energy API

braddf:
# Pull Request

## Description

The team spotted during a demo that the v1 docs still call themselves the Quartz API. This renames them to the OCF Energy API, moves support to the new address, and tidies up how the introduction reads. Three commits, separable in review: the rename, the layout and type scale, then the fonts.

image

**What changed, all v1-only**
- The docs title is now `OCF Energy API`. It was `Quartz API Documentation` — "Documentation" went too, since the page is self-evidently the docs.
- The intro line is now `Renewable generation forecasts and observed data across multiple countries.` The old wording opened with "The Quartz API provides…" and closed with "powered by Open Climate Fix", both of which just repeated the title and the sidebar logo.
- The contact link shows `support@ocf.energy` instead of `Quartz API by Open Climate Fix`, which repeated the title again and hid the actual address behind a mailto.
- `SUPPORT_EMAIL` is now `support@ocf.energy`. That constant is v1-only, so this also moves the three places it appears in the schema: the extended-history note, the access-request line, and the intraday-model 400.
- The sign-up link points at `app.ocf.energy`, the new production frontend.
- The contact URL points at `openclimatefix.org`, since `ocf.energy` redirects there and there's no separate site planned.

## Intro layout and heading scale

The introduction rendered in the same narrow column as an endpoint description, so the prose was cramped while the Server / Authentication / Client Libraries cards beside it had room to spare. Scalar has no configuration for this — checked against 1.69.0, the current release — so this rebalances the flex columns it already emits: the text takes the free space, the cards keep a fixed width.

Scalar also renders every heading level in a description identically, both `h2` and `h3` at 16px/600 in the same colour, so a nested heading read as another top-level one. Headings now have a scale, and the whole section is set from a single base size with the levels in `em`, so there is one number to change if it wants to be bigger or smaller.

`Resources` is now three headed subsections instead of a bullet list, which is what prompted the heading work.

Everything here is selected by tag name and by what a column contains, never by position or count. That is deliberate: an earlier attempt laid the sections out as cards, which needed rules keyed to each block's index and broke the moment a heading was added. Editing the description cannot break this version.

## Font updated to OCF's Matter family

The docs now use Matter XH and Matter SemiMono, self-hosted from `/static/fonts/` the same way the frontend does it, with the same three weights. Scalar's default fonts are turned off — they fetched Inter and JetBrains Mono from Google, which nothing then used.

`*.woff2` is added to `package-data`. Without it the fonts are absent from the built image and the page quietly falls back to system sans, which is how the SVGs went missing before #395.

Worth flagging separately: Matter is a commercial face. The frontend already serves it publicly, but this puts it on a second origin, so it is worth confirming the webfont licence is not domain-scoped.

The header row was an even split, so the title wrapped onto two lines while the contact link sat in a mostly empty column. It is now two thirds to one third. That override applies above 1100px only — Scalar collapses the row to one column itself through a container query, and a container query adds no specificity, so an unscoped rule would win there too and squash the two together on a phone.

## v0 remains as-is

`_custom_openapi` builds the OpenAPI schema for both apps and hardcoded a single contact block. That block is now passed in per app, so v0 keeps `Quartz API by Open Climate Fix` / `www.quartz.solar` / `support@quartz.solar` exactly as before.

The root ReDoc page still has a `Quartz API Docs` title. That's v0's, and it's left for the v0 sunset rather than changed here.

## Testing

- Unit tests and ruff pass.
- No behaviour changes — this is docs copy, CSS, an OpenAPI contact block, and a constant.
- Checked the rendered docs page on a local server, including with headings added to the description to confirm the layout holds.
- Confirmed the fonts are served as `font/woff2` and that the browser reports both families as loaded, rather than silently falling back.

## Checklist:


devsjc (approved):

braddf:
Sorry @devsjc not trying to be sneaky, just remembered I needed to set the brand font too!

